PR #193, opened 2020-12-09
Problem: Missing message sender and sequence (wip #92)

leejw51crypto:
this is work in progress


leejw51crypto:
this is already done in other prs

PR #246, opened 2020-12-28
Problem: Missing account projections (fix #98)

### LINKED ISSUE #98: Problem: Missing account projections

leejw51crypto:
Solution: add account projection
for balance, account number, used roc to fetch info

now all accounts used in blockchain are indexed and queried
and balance, account number is correct


leejw51crypto:
ok,i'll add comment

cdc-Hitesh (approved):
Cool.

PR #264, opened 2020-12-31
Problem: Missing Crossfire TxSentRank 

### LINKED ISSUE #241: Problem: Missing business logic of crossfire validator's tasks completion
For validator's tasks completion, it could be divided into the following 7 tasks and each task may in a different phase:

- [x] taskSetup
- [x] taskKeepActive ( WIP PR: #260 )
- [x] taskProposalVote (relatively low-priority)
- [x] taskNetworkUpgrade (relatively low-priority)
- [x] taskTxSentRank (WIP PR : https://github.com/crypto-com/chain-indexing/pull/264)
- [x] taskPhase1n2CommitmentRank
- [x] taksPhase3CommitmentRank

The detailed technical spec could be found [here](https://docs.google.com/presentation/d/1AIxybRm3GFqsM_HbgDpEYnAnM7HJBI1Ll_-NQ_-64mg). This task comes with #239 #240 will resolve #226 .

cdc-Hitesh:
WIP

FIxes #241


appinterface/projection/crossfire/view/crossfire_validators_stats.go
```diff
@@ -63,13 +63,42 @@ func (view *CrossfireValidatorsStats) FindBy(key string) (int64, error) {
 	return value, nil
 }
 
+func (view *CrossfireValidatorsStats) FindAllBy(key string) ([]CrossfireValidatorsStatsRow, error) {
```
foreseaz, line 66: Maybe `FindAllLike` is more appropriate?

appinterface/projection/crossfire/crossfire.go
```diff
@@ -461,9 +484,115 @@ func (projection *Crossfire) checkTaskNetworkProposalVote(
...
+
+				// Check if Participant is a Registered Crossfire Validator
+				registeredCrossfireValidator, errFindingValidator := crossfireValidatorView.FindBy(view.CrossfireValidatorIdentity{
+					MaybeOperatorAddress: &participant.OperatorAddress,
+				})
+				if errFindingValidator != nil || registeredCrossfireValidator == nil {
+					break
+				}
+
+				// Update Ranks with specific Validator
```
foreseaz, line 577: From the previous discussion, there is a hidden logic here, it needs to check if two tx count is the same and return the same rank. So that it can have multiple rank 10 and all of them will get the reward.

foreseaz (changes requested):
Great PR. And there is some missing logic.

appinterface/projection/crossfire/view/crossfire_validators_stats.go
cdc-Hitesh: Makes sense.

appinterface/projection/crossfire/crossfire.go
cdc-Hitesh: Just need to confirm this? Whether we are going with this way or not? I am waiting on the discussion to conclude on the slack thread.   I will update accordingly thanks Owen.

appinterface/projection/crossfire/crossfire.go
```diff
@@ -461,9 +485,115 @@ func (projection *Crossfire) checkTaskNetworkProposalVote(
...
+				errUpdating := crossfireValidatorView.UpdateTxSentRank(rank, participant.PrimaryAddress, participant.OperatorAddress)
+				if errUpdating != nil {
+					return fmt.Errorf("[error] Updating TX SENT Task Rank %w", errUpdating)
+				}
+				if index+1 < len(dbParticipantWithCountList) && dbParticipantWithCountList[index].Value != dbParticipantWithCountList[index+1].Value {
```
foreseaz, line 581: @cdc-Hitesh I'm a little bit confused about this logic, does it mean that `only current tx_count not equal to next tx_count` then `rank++`? If so, then the result will be `1, 2, 3, 4, 4, 4, 5` instead of `1, 2, 3, 4, 4, 4, 7` right?
I think we can keep the `rank++` for every loop and maintain a variable for `last_index_tx_count` and `last_rank`, so if the count are equal, use the `last_rank`, if not use the `rank`. Do you think this is feasible?

cdc-Hitesh: @foreseaz Thanks.
>If so, then the result will be 1, 2, 3, 4, 4, 4, 5 instead of 1, 2, 3, 4, 4, 4, 7 right?

Yes, you are correct.
Can I understand a dataset where this can happen `1, 2, 3, 4, 4, 4, 7`? Because after rank `#_4` only #_5 is supposed to come right?

appinterface/projection/crossfire/crossfire.go
```diff
@@ -58,6 +61,7 @@ func NewCrossfire(
 		competitionEndTime:       utctime.FromUnixNano(unixCompetitionEndTime),
 		adminAddress:             adminAddress, // TODO: address prefix check
 		networkUpgradeProposalID: networkUpgradeProposalID,
+		participantsListUrl:      participantsListURL,
 		rdbConn:                  rdbConn,
 		logger:                   logger,
 	}
```
foreseaz, line 67: Should add listenning event `event_usecase.TRANSACTION_CREATED`

appinterface/projection/crossfire/crossfire.go
```diff
@@ -344,25 +355,38 @@ func (projection *Crossfire) projectCrossfireValidatorView(
...
-				return fmt.Errorf("error converting ProposalID to int64: s%v", errConversion)
+				return fmt.Errorf("error converting ProposalID to int64: %v", errConversion)
 			}
 			errUpdateValidatorStats := crossfireValidatorStatsView.Set(voted_proposal_id_db_key, proposalIdAsInt64)
 
 			if errUpdateValidatorStats != nil {
-				return fmt.Errorf("error Updating ProposalID for the voter: s%v", errUpdateValidatorStats)
+				return fmt.Errorf("error Updating ProposalID for the voter: %v", errUpdateValidatorStats)
+			}
+		} else if msgTransactionCreated, ok := event.(*event_usecase.TransactionCreated); ok {
```
foreseaz, line 378: Refer to https://github.com/crypto-com/chain-indexing/blob/master/appinterface/projection/transaction/transaction.go#L74, it should be
```
if transactionCreatedEvent
```
?

appinterface/projection/crossfire/httpclient.go
```diff
@@ -23,22 +24,20 @@ func NewHTTPClient(participantsListURL string) *HTTPClient {
...
 	return &HTTPClient{
 		httpClient,
-		participantsListURL,
+		strings.TrimSuffix(participantsListURL, "/"),
 	}
 }
 
 func (client *HTTPClient) Participants() (*[]Participant, error) {
-	rawRespBody, err := client.request(client.url)
+	rawRespBody, err := client.request("participants.json")
```
foreseaz, line 32: May I know why we hardcoded a `participants.json` here? and also the current participants URL in config is still full URL right?

appinterface/projection/crossfire/httpclient.go
```diff
@@ -47,7 +46,7 @@ func (client *HTTPClient) Participants() (*[]Participant, error) {
 func (client *HTTPClient) request(method string, queryString ...string) (io.ReadCloser, error) {
 	var err error
 
-	url := client.url
+	url := client.url + "/" + method
```
foreseaz, line 49: I think there is no need to add `method` here, since it's a JSON file.

appinterface/projection/crossfire/httpclient_test.go
```diff
@@ -23,7 +23,7 @@ var _ = Describe("Crossfire Chain Stats", func() {
 			1,
 			"foo",
 			"14",
-			"https://raw.githubusercontent.com/foreseaz/random/master/participants.json",
+			"https://raw.githubusercontent.com/foreseaz/random/master/",
```
foreseaz, line 26: Maybe it's better to keep the original full URL and it's better to replace them as a whole.

appinterface/projection/crossfire/crossfire_test.go
```diff
@@ -35,11 +39,207 @@ var _ = Describe("Crossfire", func() {
 			1,
 			"foo",
 			"14",
-			"https://example.com/participants.json",
+			"https://raw.githubusercontent.com/foreseaz/random/master/",
```
foreseaz, line 42: Same above, maybe the full URL ends with JSON is better for config management.

appinterface/projection/crossfire/crossfire_test.go
```diff
@@ -117,7 +289,7 @@ var _ = Describe("Crossfire", func() {
 				1613361599000000000,
 				"tcro15grftg88l0gdw4mg9t9pwnl0pde2asjzvfpkp4",
 				"14",
-				"https://example.com/participants.json",
+				"https://raw.githubusercontent.com/foreseaz/random/master/",
```
foreseaz, line 292: Same above

foreseaz (changes requested):
Great logic, some issues need to be addressed.

appinterface/projection/crossfire/httpclient.go
cdc-Hitesh: Yes, THanks for asking this. Even I didn't want to do this, but due to the `ghttp` ginkgo unit testing limitation it had to be done. I couldn't figure out a way to keep it like previous and get the unit test to pass.
Maybe if we find out some way then we can go back to original.

appinterface/projection/crossfire/httpclient.go
cdc-Hitesh: Yes, I agree but due to https://github.com/crypto-com/chain-indexing/pull/264#discussion_r553123893 I had to do this.

appinterface/projection/crossfire/httpclient_test.go
cdc-Hitesh: https://github.com/crypto-com/chain-indexing/pull/264#discussion_r553123893

appinterface/projection/crossfire/crossfire_test.go
cdc-Hitesh: https://github.com/crypto-com/chain-indexing/pull/264#discussion_r553123893

appinterface/projection/crossfire/crossfire_test.go
cdc-Hitesh: https://github.com/crypto-com/chain-indexing/pull/264#discussion_r553123893

appinterface/projection/crossfire/crossfire.go
cdc-Hitesh: Yes yes correct.

appinterface/projection/crossfire/crossfire.go
cdc-Hitesh: Loll. Thanks for pointing

appinterface/projection/crossfire/httpclient.go
cdc-Hitesh: Update here:
Managed to run and add test case using ghttp now. Thanks. Have pushed again

foreseaz (approved):
LGTM

PR #266, opened 2021-01-02
Problem: missing PASSWORD check (fix #265)

### LINKED ISSUE #265: Problem: missing PASSWORD check
check whether password is legitimate
if not legitimate, panic

leejw51crypto:
Solution: add password check for Postgres

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,12 +63,15 @@ This will start the following docker instances on your local network when you us
 
 ### 2.4 Execute Database Migration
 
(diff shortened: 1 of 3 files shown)


leejw51crypto:
this pr is to protect servers.
because even in testing, too easy password can be rabbit hole.
also docker, i think DB_PASSWORD should be checked thoroughly.

i suggest writing to document,too.

leejw51crypto:
i'll separate checking function to other program, and activate it in pmmigrate.sh

leejw51crypto:
rebased, and in preparation of database,
if db password is not strong enough, stop further processing

calvinaco (approved):

Issue #240, opened 2020-12-24
Problem: Missing Crossfire validator projection HTTPAPI & handlers


foreseaz:
This task mainly focuses on HTTP API implementation on `/crossfire/rank` and its handlers. It will consume the data from `view_tables` created in #239 and serve the frontend.

Current draft API endpoint, params and response are:

```
https://chain.crypto.com/explorer/api/v1/crossfire/rank?order=score.desc

[
  {
    "operatorAddress": "tcrocncl15grftg88l0gdw4mg9t9pwnl0pde2asjzekz0ek",
    "consensusNodeAddress": "tcrocnclcons1xuza5new2wsfqfw64rn9s8huapgczxg5z5vct8",
    "initialDelegatorAddress": "tcro15grftg88l0gdw4mg9t9pwnl0pde2asjzvfpkp4",
    "joinedAtBlockHeight": 68238,
    "lastSignedBlockHeight": 92554,
    "moniker": "1NP_idarrib",
    "phase1n2Commits": 23394,
    "phase3Commits": 12734,
    "transactionSent": 2637,
    "taskSetup": 1,
    "taskKeepActive": 1,
    "taskProposalVote": 1,
    "taskNetworkUpgrade": 1,
    "taskTxSentRank": 27,
    "taskPhase1n2CommitmentRank": 9,
    "taksPhase3CommitmentRank": 18
  },
  ...
]
```

PR #279, opened 2021-01-08
Problem: Crossfire API Handler

cdc-Hitesh:
Reference #240

### DIFF
--- a/appinterface/projection/crossfire/crossfire.go
+++ b/appinterface/projection/crossfire/crossfire.go
@@ -25,6 +25,7 @@ type Crossfire struct {
 	Client                   *HTTPClient
 	conNodeAddressPrefix     string
 	validatorAddressPrefix   string
+	primaryAddressPrefix     string
 	phaseOneStartTime        utctime.UTCTime
 	phaseTwoStartTime        utctime.UTCTime
 	phaseThreeStartTime      utctime.UTCTime
@@ -48,6 +49,7 @@ func NewCrossfire(
 	adminAddress string,
 	networkUpgradeProposalID string,
 	participantsListURL string,
+	primaryAddressPrefix string,
 ) *Crossfire {
 	return &Crossfire{
 		Base: rdbprojectionbase.NewRDbBase(rdbConn.ToHandle(), "Crossfire"),
@@ -64,6 +66,7 @@ func NewCrossfire(
 		participantsListUrl:      participantsListURL,
 		rdbConn:                  rdbConn,
 		logger:                   logger,
+		primaryAddressPrefix:     primaryAddressPrefix,
 	}
 }
 
@@ -541,7 +544,7 @@ func (projection *Crossfire) updateTxSentCount(
 		// Only considering Pubkey address for now
 		if sender.Type == constants.TYPE_URL_PUBKEY && sender.MaybeThreshold == nil {
 			pubKey, _ := base64.StdEncoding.DecodeString(sender.Pubkeys[0])
-			primaryAddress, err := tmcosmosutils.AccountAddressFromPubKey("tcro", pubKey)
+			primaryAddress, err := tmcosmosutils.AccountAddressFromPubKey(projection.primaryAddressPrefix, pubKey)
 			if err != nil {
 				continue
 			}
--- a/appinterface/projection/crossfire/crossfire.go
+++ b/appinterface/projection/crossfire/crossfire.go
@@ -540,16 +540,21 @@ func (projection *Crossfire) updateTxSentCount(
 
 		// Only considering Pubkey address for now
 		if sender.Type == constants.TYPE_URL_PUBKEY && sender.MaybeThreshold == nil {
+			pubKey, _ := base64.StdEncoding.DecodeString(sender.Pubkeys[0])
+			primaryAddress, err := tmcosmosutils.AccountAddressFromPubKey("tcro", pubKey)
+			if err != nil {
+				continue
+			}
 
 			// Increment count for address as per PHASE
-			phaseAddressCountDbKey := phaseNumberPrefix + constants.DB_KEY_SEPARATOR + sender.Pubkeys[0]
+			phaseAddressCountDbKey := phaseNumberPrefix + constants.DB_KEY_SEPARATOR + primaryAddress
 			errIncrementing := crossfireValidatorStatsView.IncrementOne(phaseAddressCountDbKey)
 			if errIncrementing != nil {
 				return fmt.Errorf("error Phase wise tx sent count increment: %v", errIncrementing)
 			}
 
 			// Increment TOTAL count for address
-			totalAddressCountDbKey := constants.TOTAL_TX_SENT_PREFIX + constants.DB_KEY_SEPARATOR + sender.Pubkeys[0]
+			totalAddressCountDbKey := constants.TOTAL_TX_SENT_PREFIX + constants.DB_KEY_SEPARATOR + primaryAddress
 			errIncrementingTotal := crossfireValidatorStatsView.IncrementOne(totalAddressCountDbKey)
 			if errIncrementingTotal != nil {
 				return fmt.Errorf("error Incrementing tx sent count: %v", errIncrementingTotal)
--- a/infrastructure/httpapi/handlers/crossfire.go
+++ b/infrastructure/httpapi/handlers/crossfire.go
@@ -111,35 +111,28 @@ func (handler *Crossfire) getValidatorStats(operatorAddress string) (*ValidatorS
 		return nil, fmt.Errorf("[error-crossfire] Fetching Validator Stats, %v", err)
 	}
 
-	var finalValidatorStats ValidatorStats
+	var finalValidatorStats *ValidatorStats
 	for _, validatorStatRow := range validatorStatRows {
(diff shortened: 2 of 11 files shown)


infrastructure/httpapi/handlers/crossfire.go
```diff
@@ -1,9 +1,162 @@
...
+func (handler *Crossfire) getValidatorStats(operatorAddress string) (*ValidatorStats, error) {
+	// Fetch Validator Stats
+	key := fmt.Sprintf("%s%s", "%", operatorAddress)
+	validatorStatRows, err := handler.crossfireValidatorsStatsView.FindAllLike(key)
+	if err != nil {
+		return nil, fmt.Errorf("[error-crossfire] Fetching Validator Stats, %v", err)
+	}
+
+	var finalValidatorStats *ValidatorStats
+	for _, validatorStatRow := range validatorStatRows {
```
foreseaz, line 115: Smart implementation! One case may need change is, the `PHASE1_COMMIT` is updated to `PHASE_1N2_COMMIT_PREFIX`

infrastructure/httpapi/handlers/crossfire.go
```diff
@@ -1,9 +1,162 @@
...
+	crossfireValidatorsDB, err := handler.crossfireValidatorsView.List()
+	if err != nil {
+		handler.logger.Errorf("error listing validators: %v", err)
+		httpapi.InternalServerError(ctx)
+		return
+	}
+
+	var finalList = []CrossfireValidatorDetails{}
+
+	for _, participant := range *participantList {
```
foreseaz, line 75: A small suggestion, we can use a map to filter out the participated validators from  `crossfireValidatorsDB`, to avoid the O(N^2) nested loop

infrastructure/httpapi/handlers/crossfire.go
```diff
@@ -1,9 +1,162 @@
...
+		}
+	}
+
+	httpapi.Success(ctx, finalList)
+}
+
+func (handler *Crossfire) getValidatorStats(operatorAddress string) (*ValidatorStats, error) {
+	// Fetch Validator Stats
+	key := fmt.Sprintf("%s%s", "%", operatorAddress)
+	validatorStatRows, err := handler.crossfireValidatorsStatsView.FindAllLike(key)
```
foreseaz, line 109: For this one, I'm not sure if the `FindAllLike` here will work or not. Since our SQL is 
```
"key LIKE ?", fmt.Sprint(key, "%")
=> SQL 
"key LIKE tcroxxxxxxx%"
```
but we are now storing the operatorAddress on the right-hand side of `:`, so the LIKE SQL should be like following right?
```
"key LIKE %tcroxxxxxxx" // Finds any values that end with "tcroxxxxxx"
```

Then we may need to tweak `FindAllLike` a little bit.

infrastructure/httpapi/handlers/crossfire.go
```diff
@@ -1,9 +1,162 @@
...
+	commitCountPhase1 *int64
+	commitCountPhase2 *int64
+	commitCountPhase3 *int64
+}
+
+// CrossfireValidatorDetails response object
+type CrossfireValidatorDetails struct {
+	operatorAddress            string
+	initialDelegatorAddress    string
+	moniker                    string
```
foreseaz, line 153: @cdc-Hitesh could you also add following fields for return:

```
- join height
- join block time
- tx sent count
- phase1n2 commitment count
- phase3 commitment count
```

Thanks!

foreseaz (changes requested):
Great implementation! Some issues may need updates.

infrastructure/httpapi/handlers/crossfire.go
foreseaz: If you consider going for a Map to filter, a quick reference could be [here](https://github.com/crypto-com/chain-indexing/blob/1ad76f851e1394547fe3375bd36e471abc7a5d63/appinterface/projection/crossfire/crossfire.go#L569-L590)

cdc-Hitesh: Thanks will update accordingly

infrastructure/httpapi/handlers/crossfire.go
cdc-Hitesh: Cool. Noted

infrastructure/httpapi/handlers/crossfire.go
cdc-Hitesh: It shall be `%tcro...%`, As my key is : `fmt.Sprintf("%s%s", "%", operatorAddress)`

infrastructure/httpapi/handlers/crossfire.go
cdc-Hitesh: Cool noted

cdc-Hitesh:
Marking this PR as Draft in progress

infrastructure/httpapi/handlers/crossfire.go
foreseaz: Oh, that should be fine~

infrastructure/httpapi/handlers/crossfire.go
cdc-Hitesh: Btw, DOes that mean `PHASE_1N2_COMMIT_PREFIX ` it consists of both phase commit together ?

infrastructure/httpapi/handlers/crossfire.go
cdc-Hitesh: @foreseaz  Oh It's already there under `stats`  of type `ValidatorStats`. Do you mean something else?

infrastructure/httpapi/handlers/crossfire.go
cdc-Hitesh: @foreseaz 
I wonder the `getValidatorStats` count can never be 40000 times as there is a `IF(....)` check before it selects from DB. 
So there shall be exactly `200` calls ( Only for the matching ones )

However, Map is a fantastic approach, but wouldn't optimise here much. In case me understanding something wrong let me know

infrastructure/httpapi/handlers/crossfire.go
foreseaz: yes, it consists of both phase commit together

infrastructure/httpapi/handlers/crossfire.go
foreseaz: Oh sorry, haven't seen that one. Should be fine! Thanks

infrastructure/httpapi/handlers/crossfire.go
foreseaz: Sure, the `if` check can ensure it as `200` calls. But it will be better to avoid the nested loop, in case future adding more logic in the loop I think.

infrastructure/httpapi/handlers/crossfire.go
```diff
@@ -1,9 +1,162 @@
...
+		return
+	}
+
+	var finalList = []CrossfireValidatorDetails{}
+
+	for _, participant := range *participantList {
+		for _, dbValidator := range crossfireValidatorsDB {
+
+			// For each matching participant create a return object
+			if dbValidator.OperatorAddress == participant.OperatorAddress && dbValidator.InitialDelegatorAddress == participant.PrimaryAddress {
```
foreseaz, line 79: For the `finalList`, so if there no `dbValidator.OperatorAddress` in the table, then it will return as empty array? As previous discussion, it's better to return the default value of the participants with the same structure.

As your response defined in the struct, should be like:
```
{
	operatorAddress.    FROM_PARTICIPANT
	initialDelegatorAddress    FROM_PARTICIPANT
	moniker                    FROM_PARTICIPANT
	taskSetup                  Incompleted
	taskKeepActive             Incompleted
	taskProposalVote           Incompleted
	taskNetworkUpgrade        Incompleted
	rankPhase1n2CommitmentRank 0
	rankPhase3CommitmentRank   0
	rankTxSentRank             0
	stats                     default all 0 ValidatorStats
}
```

infrastructure/httpapi/handlers/crossfire.go
cdc-Hitesh: Cool.

infrastructure/httpapi/handlers/crossfire.go
cdc-Hitesh: Yes. COrrect. You mentioned to keep it this way. I will push it soon.

cdc-Hitesh:
Marking this PR as ready for review.

Have to disable a few test case. Looking for correct addresses and pubkey was little time taking. WIll be adding soon.

appinterface/projection/crossfire/crossfire.go
```diff
@@ -64,6 +66,7 @@ func NewCrossfire(
 		participantsListUrl:      participantsListURL,
 		rdbConn:                  rdbConn,
 		logger:                   logger,
+		primaryAddressPrefix:     primaryAddressPrefix,
```
foreseaz, line 69: This line should better move to the line under `validatorAddressPrefix` to keep consistent with struct

foreseaz (approved):
LGTM

PR #280, opened 2021-01-09
Problem: Missing projection toggle configuration

### LINKED ISSUE #188: Problem: Missing projection toggle configuration
in the config, we can specify on/off for each projections.
by default, all is on

in case, new event is added, we can run, another chain-indexing server which updates only
new specific projection.
in that way, we can maintain service without stopping.

after new event table, catch up with the latest, it can be turned off.

calvinlauco:
Solution: (Fix #188) Add config


projection/projections.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/crypto-com/chain-indexing/projection/transaction"
+	"github.com/crypto-com/chain-indexing/projection/validator"
+	"github.com/crypto-com/chain-indexing/projection/validatorstats"
+)
+
+func InitProjection(name string, params InitParams) projection_entity.Projection {
+	switch name {
+	case "Account":
+		return account.NewAccount(params.Logger, params.RdbConn, params.CosmosAppClient, params.BaseDenom)
+	case "AccountMessage":
```
cdc-Hitesh, line 23: Is it any benefit to extract cases MATCHERS out to a constant file?

cmd/chain-indexing/projections.go
```diff
@@ -1,38 +1,37 @@
...
-		account_message.NewAccountMessage(logger, rdbConn),
-		account.NewAccount(logger, rdbConn, cosmosAppClient, config.Blockchain.BaseDenom),
 
-		// register more projections here
+	projections := make([]projection_entity.Projection, 0, len(config.Projection.Enables))
+	initParams := projection.InitParams{
+		Logger:                logger,
+		RdbConn:               rdbConn,
+		CosmosAppClient:       cosmosAppClient,
+		ConsNodeAddressPrefix: config.Blockchain.ConNodeAddressPrefix,
```
cdc-Hitesh, line 25: I think for a longer run, We shall instead of passing Address Prefixes we can pass `Network TYPE `value instead and let a central package manage address prefixes according to the network type.

Nothing major but will help.

cmd/chain-indexing/projections.go
```diff
@@ -1,38 +1,37 @@
...
+		BaseDenom:             config.Blockchain.BaseDenom,
+	}
+	for _, projectionName := range config.Projection.Enables {
+		projections = append(projections, projection.InitProjection(
+			projectionName, initParams,
+		))
 	}
+
+	logger.Infof("Enabled the follow projections: [%s]", strings.Join(config.Projection.Enables, ", "))
```
cdc-Hitesh, line 34: I think we should even handle for ZERO projections enabled. TOtally unlikely though 😆 

Maybe running Chain indexing in NO-PROJECTION Mode

cdc-Hitesh (commented):
Left some comments/thougths

cmd/chain-indexing/projections.go
calvinlauco: Yes, passing no projection should not break though. It will display an empty array here.

cmd/chain-indexing/projections.go
calvinlauco: I tend to leave this option to be a config because in the long run we can serve not only Crypto.com Chain but can be used by another Cosmos SDK based chain

projection/projections.go
calvinlauco: Um, could do. As long as the constant and init is in the same file. The design is to make adding new projection does not need to amend too many files (and may miss some if too many)

PR #289, opened 2021-01-12
Feat: amend activeness logic for activeness

### LINKED ISSUE #284: Problem: Crossfire activeness task logic need amendment
Primary Node must be setup on phase 1. And any activeness task should check task_phase_1_node_setup. Including:

- Entry level  always up task (during phase2)
- MVP up time task (during phase2)
- MVP up time task (during phase3)

foreseaz:
Disable taskActive and commitment ranking for the validator who missed taskSetup, resolved #284

### DIFF
--- a/appinterface/projection/crossfire/crossfire.go
+++ b/appinterface/projection/crossfire/crossfire.go
@@ -472,6 +472,12 @@ func (projection *Crossfire) checkTaskKeepActive(
 	}
 
 	for _, validator := range validators {
+		// If validator missed the task 1, it cannot complete the active task
+		if validator.TaskPhase1NodeSetup == constants.MISSED {
+			projection.logger.Debugf("Won't handle task keepActive for %s because it missed taskSetup", validator.OperatorAddress)
+			continue
+		}
+
 		key := constants.ValidatorCommitmentKey(validator.OperatorAddress, constants.PHASE_2_COMMIT_PREFIX)
 		validatorPhase2CommitCount, err := crossfireValidatorsStatsView.FindBy(key)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


appinterface/projection/crossfire/crossfire.go
```diff
@@ -472,6 +472,12 @@ func (projection *Crossfire) checkTaskKeepActive(
 	}
 
 	for _, validator := range validators {
+		// If validator missed the task 1, it cannot complete the active task
+		if validator.TaskPhase1NodeSetup == constants.MISSED {
+			projection.logger.Debugf("Won't handle task keepActive for %s because it missed taskSetup", validator.OperatorAddress)
```
cdc-Hitesh, line 477: This Log goes in the chain-indexing logs. 
Is that behaviour fine?

appinterface/projection/crossfire/crossfire.go
```diff
@@ -581,9 +587,28 @@ func (projection *Crossfire) computeCommitmentRank(
...
+		// check whether validator missed taskSetup or not
+		validatorRow, err := crossfireValidatorView.FindBy(view.CrossfireValidatorIdentity{
+			MaybeOperatorAddress: &statsOperatorAddress,
+		})
+		if errors.Is(err, rdb.ErrNoRows) {
+			// no validator found by tendermint address
+			continue
+		}
+		if err != nil {
+			return fmt.Errorf("error getting validator with operator addr %s %v", statsOperatorAddress, err)
```
cdc-Hitesh, line 604: This will revert the whole process and will stop. Is that intended?

cdc-Hitesh (commented):
Left some comments. Otherwise looks good

appinterface/projection/crossfire/crossfire.go
foreseaz: Yes, should be~ if the error is non-norow error, then it's a big problem and should be fixed by us

appinterface/projection/crossfire/crossfire.go
foreseaz: I think it should be fine

cdc-Hitesh (approved):

PR #301, opened 2021-01-14
Feature/tx rank bug

calvinaco:
1. Tx ranking is based on unfiltered list (include all senders, not only participants), hence the ranking increment logic is incorrect
    - e.g. Tranaction count table:
`A:5`, `B:5`, `C:5`, `D:2`, `E:2`; where A, D, E are participants;
old logic rank will result:
`A: 1`, `B: N/A`, `C: N/A`, `D: 1`, `E: 1`
because after checking A, the logic will think the next one in the list, i.e. B has the same tx count. so the rank is not incremented; But actually B is not a participant When it reaches D, the rank is not incremented. Hence the error

2.  The tx rank update logic assume each participant operator address and primary address has relationship
(i.e. `primary address` <----`bech32 de/encode` ---> `operator address`), but it does not always hold:
In particular the update SQL use a  `where condition` which checks both addresses from participant list should match with the DB record

3. Network upgrade use FindBy to get the upgrade time, which has default value of 0 when no row is found, hence causing logic error


appinterface/projection/crossfire/crossfire.go
```diff
@@ -691,19 +692,33 @@ func (projection *Crossfire) computeTxSentRank(
...
 
+	// Filter only participants from tx senders list
+	dbParticipantWithCountList := make([]view.CrossfireValidatorsStatsRow, 0)
+	for _, dbTxSender := range dbTxSendersWithCountList {
+		dbParticipantPrimaryAddress := strings.Split(dbTxSender.Key, constants.DB_KEY_SEPARATOR)[1]
+		if _, ok := participantsMap[dbParticipantPrimaryAddress]; ok {
+			dbParticipantWithCountList = append(dbParticipantWithCountList, dbTxSender)
+		}
+	}
+
```
calvinaco, line 721: @cdc-Hitesh Found a bug regarding the ranking logic, please help to verify if this has fixed it.

calvinaco (commented):
@cdc-Hitesh

appinterface/projection/crossfire/crossfire.go
cdc-Hitesh: Looks good.

appinterface/projection/crossfire/view/crossfire_validators.go
```diff
@@ -419,16 +419,14 @@ UpdateTxSentRank : Updates the participating validator rank
...
 	operatorAddress string,
 ) error {
 	sql, sqlArgs, err := view.rdb.StmtBuilder.Update(
 		TABLE_NAME,
 	).Set(
 		"rank_task_highest_tx_sent", rank,
 	).Where(
-		"operator_address = ? AND initial_delegator_address = ?",
-		operatorAddress, primaryAddress,
+		"operator_address = ?", operatorAddress,
```
cdc-Hitesh, line 429: Fantastic find.

cdc-Hitesh (approved):

cdc-Hitesh:
Shall I merge it @calvinaco

appinterface/projection/crossfire/crossfire.go
```diff
@@ -195,21 +195,19 @@ func (projection *Crossfire) handleBlockCreatedEvent(
...
 		err := crossfireChainStatsView.IncrementOne(constants.PHASE_2_BLOCK_COUNT)
 		if err != nil {
 			return fmt.Errorf("[Crossfire] error increment phase2 block count")
 		}
-		projection.profile("end incrementing phase 2 block count")
 	} else if blockTime.AfterOrEqual(projection.phaseThreeStartTime) && blockTime.Before(projection.competitionEndTime) {
 		// check the keep active task, throttling with every 10 blocks
+		// keep active task is only counting commitments during phase 2, so checking it in phase 3
+		// guarantees correctness
 		if blockHeight%10 == 0 {
```
cdc-Hitesh, line 211: For the last `%10` Number of the block, It may not check to keep active?
For eg: 
Let's say: The `blockheight` at `competitionEndTime` happens to be `99`
Now our logic will last execute `checkTaskKeepActive ` at `90` .
Which means for `(99-90 = 9)` Blocks we would have missed the task completion checking.

Shall we remove `blockHeight%10 == 0` check to mitigate this?
@calvinaco

calvinaco: The keep active task concerns of up time during phase 2 only; while this check is done during phase 3. So at the time it is executed the keep alive data are fixed.

So there shouldn't be problem. In fact we may further extend to longer time interval for optimizing efficiency.

cdc-Hitesh: Ahh. True.. . .The underlying function only deals with phase 2 values. Cool

PR #325, opened 2021-01-21
Add Crossfire jackpot winner finder

calvinaco:
```
go run cmd/find-jackpot-winners/* [block-height] [output-file]
```


cdc-Hitesh (approved):

allthatjazzleo:
should return list of `crossfire.Participant`

PR #336, opened 2021-01-27
Optimisation crossfire insert

cdc-Hitesh:


cdc-Hitesh:
I think this approach can even be replicated in the validator commitment count insertion as well, instead of incrementing by one we increment cumulatively block wise. @calvinaco

appinterface/projection/crossfire/view/crossfire_validators_stats.go
```diff
@@ -136,6 +136,35 @@ func (view *CrossfireValidatorsStats) IncrementAllByOne(keys []string) error {
...
+func (view *CrossfireValidatorsStats) IncrementAllByCumulativeCount(keyCountMap map[string]int64) error {
+	stmtBuilder := view.rdbHandle.StmtBuilder.Insert(
+		CROSSFIRE_VALIDATOR_STATS_VIEW_TABLENAME,
+	).Columns(
+		"key", "value",
+	)
+
+	for key, value := range keyCountMap {
+		stmtBuilder = stmtBuilder.Values(key, value)
+	}
```
calvinaco, line 148: One suggestion here, there is a limitation of parameters allowed in a single query. So it will be good to limit the number of rows to insert on each execution.

an example can be found here:
https://github.com/crypto-com/chain-indexing/blob/master/projection/account_transaction/view/account_transactions.go#L33

calvinaco (changes requested):

appinterface/projection/crossfire/view/crossfire_validators_stats.go
cdc-Hitesh: Thanks @calvinaco I just pushed the limitation support.

cdc-Hitesh: We might need to do it for IncrementAllByOne method also?

PR #337, opened 2021-01-27
Optimisation FindAllLike style

cdc-Hitesh:
This is just a removal of dirty LIKE command done in crossfire HTTP handler.


calvinaco:
Instead of tweaking behaviour of `FindAllLike`, should we introduce a `FindAll` instead which does an exact match for better optimization?

PR #358, opened 2021-02-10
Problem: slash and jail event parsing not working properly

### LINKED ISSUE #357: Problem: slash and jail event parsing not working properly

calvinlauco:
Solution: (Fix #357) update logic to cater for missing signature and double sign


allthatjazzleo (approved):
LGTM

PR #375, opened 2021-04-11
Problem: Missing average block time

### LINKED ISSUE #374: Problem: Missing average block time

calvinaco:
Solution: (Fix #374) Add block time projection and API

### DIFF
--- a/projection/chainstats/chainstats.go
+++ b/projection/chainstats/chainstats.go
@@ -1,0 +1,158 @@
+package chainstats
+
+import (
+	"fmt"
+	"math/big"
+	"strconv"
+
+	"github.com/crypto-com/chain-indexing/internal/utctime"
+	"github.com/pkg/errors"
+
+	"github.com/crypto-com/chain-indexing/appinterface/projection/rdbprojectionbase"
+	"github.com/crypto-com/chain-indexing/appinterface/rdb"
+	event_entity "github.com/crypto-com/chain-indexing/entity/event"
+	entity_projection "github.com/crypto-com/chain-indexing/entity/projection"
+	applogger "github.com/crypto-com/chain-indexing/internal/logger"
+	"github.com/crypto-com/chain-indexing/projection/chainstats/view"
+	event_usecase "github.com/crypto-com/chain-indexing/usecase/event"
+)
+
+var _ entity_projection.Projection = &ChainStats{}
+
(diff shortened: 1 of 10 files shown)


projection/chainstats/chainstats.go
```diff
@@ -0,0 +1,158 @@
...
+	}()
+
+	rdbTxHandle := rdbTx.ToHandle()
+	chainStatsView := view.NewChainStats(rdbTxHandle)
+
+	event := events[0]
+	var maybeGenesisBlockTime *int64
+	if genesisCreatedEvent, ok := event.(*event_usecase.GenesisCreated); ok {
+		genesisBlockTime, parseErr := utctime.Parse(
+			"2006-01-02T15:04:05.000000Z", genesisCreatedEvent.Genesis.GenesisTime,
```
cdc-Hitesh, line 106: Which genesis timestamp it is?
Should it be CROs genesis TS?

projection/chainstats/chainstats.go
```diff
@@ -0,0 +1,158 @@
...
+	if err = rdbTx.Commit(); err != nil {
+		return fmt.Errorf("error committing changes: %v", err)
+	}
+	committed = true
+
+	if maybeGenesisBlockTime != nil {
+		projection.maybeGenesisBlockTime = maybeGenesisBlockTime
+		projection.maybeTotalBlockCount = big.NewInt(0)
+	} else {
+		projection.maybeTotalBlockCount = new(big.Int).Add(projection.maybeTotalBlockCount, big.NewInt(1))
```
cdc-Hitesh, line 155: any chance `projection.maybeTotalBlockCount` could be `nil`, if yes will it misbehave here while doing `.Add`

cdc-Hitesh (changes requested):

projection/chainstats/chainstats.go
calvinaco: Not really. Here we do not assume any chain. This time is the layout string as specified in https://golang.org/pkg/time/#Parse

projection/chainstats/chainstats.go
calvinaco: Good question, I have specifically handled the initialization and this value is set on two cases
- OnInit (for cases where some blocks is processed)
- If not, it is first initialized and after processing the genesis block it will be set.

So there's no way it is nil

PR #413, opened 2021-05-16
Problem: Missing NFT projection and APIs

### LINKED ISSUE #412: Problem: Missing NFT projection and APIs

calvinaco:
Solution: (Fix #412) Add NFT projection and APIs

### DIFF
--- a/config/config.mainnet.toml
+++ b/config/config.mainnet.toml
@@ -75,4 +75,6 @@ enables = [
     "Transaction",
     "Validator",
     "ValidatorStats",
+    "NFT",
+#    "CryptoComNFT",
 ]
--- a/migrations/20210514174506_view_nft_token_transfers.up.sql
+++ b/migrations/20210514174506_view_nft_token_transfers.up.sql
@@ -1,0 +1,12 @@
+CREATE TABLE view_nft_token_transfers (
+    id BIGSERIAL,
+    denom_id VARCHAR NOT NULL,
+    token_id VARCHAR NOT NULL,
+    drop VARCHAR NULL,
+    block_height BIGINT NOT NULL,
+    transaction_hash VARCHAR NOT NULL,
+    sender VARCHAR NOT NULL,
+    recipient VARCHAR NOT NULL,
+    transferred_at BIGINT NOT NULL,
+    PRIMARY KEY(id)
+)
--- a/migrations/20210514174506_view_nft_token_transfers.up.sql
+++ b/migrations/20210514174506_view_nft_token_transfers.up.sql
@@ -2,7 +2,6 @@ CREATE TABLE view_nft_token_transfers (
     id BIGSERIAL,
     denom_id VARCHAR NOT NULL,
     token_id VARCHAR NOT NULL,
-    drop VARCHAR NULL,
     block_height BIGINT NOT NULL,
     transaction_hash VARCHAR NOT NULL,
     sender VARCHAR NOT NULL,
--- a/projection/nft/nft.go
+++ b/projection/nft/nft.go
@@ -183,7 +183,12 @@ func (nft *NFT) HandleEvents(height int64, events []event_entity.Event) error {
 				LastEditedAt: blockTime,
 			}
 
-			if updateTokenErr := nft.updateToken(tokensView, tokenRow, tokensTotalView); updateTokenErr != nil {
+			if updateTokenErr := nft.updateToken(
+				tokensView,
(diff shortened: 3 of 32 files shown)


migrations/20210514174506_view_nft_token_transfers.up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE view_nft_token_transfers (
+    id BIGSERIAL,
```
cdc-Hitesh, line 2: Does it make sense to store the `view_nft_tokens` row ID here in this table?

projection/nft/nft.go
```diff
@@ -0,0 +1,321 @@
...
+		fmt.Sprintf("-:-:-:%s", tokenRow.Owner),
+		fmt.Sprintf("%s:-:-:%s", tokenRow.DenomId, tokenRow.Owner),
+		fmt.Sprintf("-:%s:-:%s", tokenRow.Drop, tokenRow.Owner),
+		fmt.Sprintf("-:-:%s:%s", tokenRow.Minter, tokenRow.Owner),
+		fmt.Sprintf("%s:%s:-:%s", tokenRow.DenomId, tokenRow.Drop, tokenRow.Owner),
+		fmt.Sprintf("%s:-:%s:%s", tokenRow.DenomId, tokenRow.Minter, tokenRow.Owner),
+		fmt.Sprintf("-:%s:%s:%s", tokenRow.Drop, tokenRow.Minter, tokenRow.Owner),
+		fmt.Sprintf("%s:%s:%s:%s", tokenRow.DenomId, tokenRow.Drop, tokenRow.Minter, tokenRow.Owner),
+	}, 1); incrementErr != nil {
+		return fmt.Errorf("error decrementing NFT token total: %v", incrementErr)
```
cdc-Hitesh, line 289:
```suggestion
		return fmt.Errorf("error incrementing NFT token total: %v", incrementErr)
```

projection/nft/nft.go
```diff
@@ -0,0 +1,321 @@
...
+	committed = true
+	return nil
+}
+
+func (nft *NFT) updateToken(tokensView *view.Tokens, tokenRow view.TokenRow, tokensTotalView *view.TokensTotal) error {
+	if updateTokenErr := tokensView.Update(tokenRow); updateTokenErr != nil {
+		return fmt.Errorf("error updating NFT token row: %v", updateTokenErr)
+	}
+
+	if decrementErr := tokensTotalView.DecrementAll([]string{
```
cdc-Hitesh, line 267: What's happening here, we are decrementing and then incrementing `tokensTotalView` by 1.
Sorry, Could you brief me?

config/config.mainnet.toml
```diff
@@ -75,4 +75,6 @@ enables = [
     "Transaction",
     "Validator",
     "ValidatorStats",
+    "NFT",
+#    "CryptoComNFT",
```
cdc-Hitesh, line 79: CryptoComNFT will be enabled when?

cdc-Hitesh (commented):
Fantastic work.
Left a few doubts and comments.

migrations/20210514174506_view_nft_token_transfers.up.sql
calvinaco: This id is an auto-incrementing id of the `view_nft_token_transfers`, it has nothing to do with `view_nft_tokens`.

projection/nft/nft.go
calvinaco: Good catch, it is trying to update previous owner's entry then increment the new one. Code updated.

config/config.mainnet.toml
calvinaco: "CryptoComNFT" is dedicated for Crypto.com based NFT. So by default it is disabled. If interested party are running Crypto.com compatible NFT formats they can disable "NFT" and enable it instead.

calvinaco: The logic's here
https://github.com/crypto-com/chain-indexing/pull/413/files#diff-36233cad5c04d2ed21a0945e0bb73cadb627342617e62ed07ca27184b8bf1d84R48

migrations/20210514174506_view_nft_token_transfers.up.sql
cdc-Hitesh: No, I mean just something like a foreign key. Though it increases complexity.

PR #446, opened 2021-07-22
fix: migrate package conflict

miraclesu:
When install migrate with command ` go get -tags 'postgres' github.com/golang-migrate/migrate/cmd/migrate`, will encounter the following error:
```
go get: module github.com/golang-migrate/migrate@upgrade found (v3.5.4+incompatible), but does not contain package github.com/golang-migrate/migrate/cmd/migrate
```


calvinaco (approved):
Looks good to me! Thanks for the PR!

PR #451, opened 2021-08-01
Add Dockerfile

### LINKED ISSUE #155: Problem: Missing Dockerfile

miraclesu:
resolves #155 
includes PR #446

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM golang:1.16
+
+WORKDIR /app
+
+COPY ./bin/migrate.linux /go/bin/migrate
+COPY ./bin/chain-indexing.linux ./chain-indexing
+COPY ./migrations ./migrations
+COPY ./pgmigrate.sh ./migrate
+
+EXPOSE 28857
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,5 +18,3 @@ WORKDIR /app
 RUN cp "${GOPATH}/src/github.com/crypto-com/chain-indexing/chain-indexing" .
 COPY ./migrations ./migrations
 COPY ./pgmigrate.sh ./migrate
-
-EXPOSE 8080
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,24 @@
+FROM golang:1.16 AS mod_cacher
+
+WORKDIR /app
+
+COPY go.mod go.sum /app/
+RUN go mod download
+
+FROM golang:1.16 AS builder
+
+COPY --from=mod_cacher $GOPATH/pkg/mod $GOPATH/pkg/mod
+COPY . "${GOPATH}/src/github.com/crypto-com/chain-indexing"
+RUN cd "${GOPATH}/src/github.com/crypto-com/chain-indexing" \
+    && make chain-indexing \
+    && go get -tags 'postgres' github.com/golang-migrate/migrate/v4/cmd/migrate
+
 FROM golang:1.16
 
 WORKDIR /app
 
-COPY ./bin/migrate.linux /go/bin/migrate
-COPY ./bin/chain-indexing.linux ./chain-indexing
+COPY --from=builder "${GOPATH}/bin/migrate" "${GOPATH}/bin/"
+COPY --from=builder "${GOPATH}/src/github.com/crypto-com/chain-indexing/chain-indexing" .
 COPY ./migrations ./migrations
 COPY ./pgmigrate.sh ./migrate
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ WORKDIR /app
 COPY go.mod go.sum /app/
 RUN go mod download
 
-FROM golang:1.16 AS builder
+FROM golang:1.16
 
 COPY --from=mod_cacher $GOPATH/pkg/mod $GOPATH/pkg/mod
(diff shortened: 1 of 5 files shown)


cdc-Hitesh (approved):

davcrypto:
@miraclesu Not nice to copy complied executable to the container directly. Suggest to complies within the docker builder and execute it once docker run.

miraclesu:
@davcrypto @cdc-Hitesh updated, please check again, thanks

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+FROM golang:1.16
+
+WORKDIR /app
+
+COPY --from=builder "${GOPATH}/bin/migrate" "${GOPATH}/bin/"
+COPY --from=builder "${GOPATH}/src/github.com/crypto-com/chain-indexing/chain-indexing" .
+COPY ./migrations ./migrations
+COPY ./pgmigrate.sh ./migrate
+
+EXPOSE 28857
```
calvinaco, line 25: I am not sure about the common practice, but should we make this port exposure to be an runtime ENV? @davcrypto  maybe can comment.

At the bottom line, since our configuration are using `8080` as the default, it will be desire-able to use it here.

calvinaco (changes requested):
Thanks @miraclesu again for your contribution. With docker file it will help us to create more reproducible build.

I have left a comment below on the exposed port.

On the other hand, I have created this task as an extension to run CI build test for the Dockerfile to make sure any code changes do not break it. Feel free to take it or our team will take it later.
https://github.com/crypto-com/chain-indexing/issues/460

Dockerfile
davcrypto: Or we can do `-p` to specify the port(s) to be exposed in `docker run` command @calvinaco

davcrypto: rather in here

Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM golang:1.16 AS mod_cacher
+
+WORKDIR /app
+
+COPY go.mod go.sum /app/
+RUN go mod download
+
+FROM golang:1.16 AS builder
```
davcrypto, line 8: @miraclesu may i know the reason you would like to use separated images for go mod, executable builder and runner?

davcrypto (changes requested):
@miraclesu Thanks for your contribution again. I left some comments below and you may take a look.

Makefile
```diff
@@ -7,4 +7,9 @@ check-password-strength:
 install:
 	go install ./cmd/chain-indexing/
 migrate:
-	./pgmigrate.sh -- -verbose up	
+	./pgmigrate.sh -- -verbose up
+
+.PHONY: docker
+docker:
+	cp .gitignore .dockerignore && echo "**/.git" >> .dockerignore
```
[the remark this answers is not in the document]
miraclesu: We only need to copy the code & config to docker image, ignore the useless files.
Files ignored by git are also ignored by docker, so reuse the `.gitignore` file, or we need to maintain `.dockerignore` file
https://docs.docker.com/engine/reference/builder/#dockerignore-file

Dockerfile
miraclesu: Generally, the changes of the mod file are relatively few compared to the code.
So, we can cache the mod packages instead of re-downloading the mod every time `docker run` after changes the code.

Makefile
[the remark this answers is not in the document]
davcrypto: Understood but not all the file ignored git should be ignored e.g. config/config.toml. So may be we can have a dedicated .dockerignore rather than coping each time

Dockerfile
```diff
@@ -5,21 +5,18 @@ WORKDIR /app
...
 WORKDIR /app
 
-COPY --from=builder "${GOPATH}/bin/migrate" "${GOPATH}/bin/"
-COPY --from=builder "${GOPATH}/src/github.com/crypto-com/chain-indexing/chain-indexing" .
+RUN cp "${GOPATH}/src/github.com/crypto-com/chain-indexing/chain-indexing" .
 COPY ./migrations ./migrations
 COPY ./pgmigrate.sh ./migrate
 
-EXPOSE 28857
+EXPOSE 8080
```
davcrypto, line 22: i mean i think we don't need to specify the port here to avoid mis-alignment i.e. in case the port in config is changed. Can just left flexibility while developer do `docker run `and they can specify on `-p` flag

davcrypto (changes requested):
@miraclesu thanks for the change but some misunderstanding on the concern. Further comment may need to follow up

Dockerfile
davcrypto: Understood. Can try in this way first.

Makefile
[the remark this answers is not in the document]
miraclesu: @davcrypto I've added `.dockerignore`, we mount config to docker with docker run, so we don't care config files, we can easily modify the configuration without having to rebuild images every time the configuration is changed.
```
docker run \
    -v `pwd`/config:/app/config --read-only \
    -p 28857:28857 \
    --env DB_PASSWORD=your_postgresql_password \
    chain-indexing /app/chain-indexing
```

Dockerfile
miraclesu: remove it

davcrypto (approved):

calvinaco (approved):

PR #463, opened 2021-08-12
Problem: Missing estimated APY based on recent n blocks

### LINKED ISSUE #452: Problem: Missing estimated APY based on recent n blocks
Right now the validator APY is estimated using the average block time of the chain starting from genesis time. However, this may be inaccurate as the block time of the network may vary from time to time.

A better estimation would be use the recent n block average block time instead.

ysong42:
Solution: Fixed #452. Calculate averageBlockTime using n latest blocks.

### DIFF
--- a/infrastructure/httpapi/handlers/validators.go
+++ b/infrastructure/httpapi/handlers/validators.go
@@ -18,10 +18,13 @@ import (
 	"github.com/crypto-com/chain-indexing/infrastructure/httpapi"
 	applogger "github.com/crypto-com/chain-indexing/internal/logger"
 	"github.com/crypto-com/chain-indexing/internal/primptr"
+	block_view "github.com/crypto-com/chain-indexing/projection/block/view"
 	chainstats_view "github.com/crypto-com/chain-indexing/projection/chainstats/view"
 	"github.com/crypto-com/chain-indexing/projection/validator/constants"
 	validator_view "github.com/crypto-com/chain-indexing/projection/validator/view"
 )
+
+const mRecentDays int64 = 7
 
 type Validators struct {
 	logger applogger.Logger
@@ -34,6 +37,7 @@ type Validators struct {
 	validatorsView          *validator_view.Validators
 	validatorActivitiesView *validator_view.ValidatorActivities
 	chainStatsView          *chainstats_view.ChainStats
+	blockView               *block_view.Blocks
 
 	globalAPY              *big.Float
 	globalAPYLastUpdatedAt time.Time
@@ -60,6 +64,7 @@ func NewValidators(
 		validator_view.NewValidators(rdbHandle),
 		validator_view.NewValidatorActivities(rdbHandle),
 		chainstats_view.NewChainStats(rdbHandle),
+		block_view.NewBlocks(rdbHandle),
 
 		nil,
 		time.Unix(int64(0), int64(0)),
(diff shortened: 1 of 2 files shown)


infrastructure/httpapi/handlers/validators.go
```diff
@@ -298,6 +303,55 @@ func (handler *Validators) getAverageBlockTime() (*big.Float, error) {
 		big.NewFloat(1000),
 	)
 
+	// In ideal case, we calculate the average block time with recent N most blocks (blocks in M recent days).
+	// N = M * 24 * 3600 / averageBlockTime
+
+	nRecentBlocksInFloat := new(big.Float).Quo(
```
davcrypto, line 309: should N be static?

ysong42: Thanks for the advice! Now it is changed to be static~

davcrypto: Thanks!

davcrypto: And i guess we can remove the existing averageBlockTime calculation part as it seems no longer needed anymore, the part that using total block time.

infrastructure/httpapi/handlers/validators.go
```diff
@@ -298,6 +311,47 @@ func (handler *Validators) getAverageBlockTime() (*big.Float, error) {
 		big.NewFloat(1000),
 	)
 
+	// totalBlockCount > nRecentBlocks: then re-calculate averageBlockTime
+	// totalBlockCount <= nRecentBlocks: use the above calculated averageBlockTime
+	nRecentBlocks := big.NewInt(nRecentBlocksInInt)
+	if totalBlockCount.Cmp(nRecentBlocks) == 1 {
```
davcrypto, line 317: Can we use a bool var to name this condition and use in here? like isTotalBlockCountExcess = (totalBlockCount.Cmp(nRecentBlocks) == 1)

infrastructure/httpapi/handlers/validators.go
```diff
@@ -298,6 +311,47 @@ func (handler *Validators) getAverageBlockTime() (*big.Float, error) {
...
+		averageBlockTimeMilliSecond = new(big.Float).Quo(
+			nRecentBlocksTotalTimeMilliSecond,
+			new(big.Float).SetInt(nRecentBlocks),
+		)
+		averageBlockTime = new(big.Float).Quo(
+			averageBlockTimeMilliSecond,
+			big.NewFloat(1000),
+		)
+
+	}
```
davcrypto, line 353: we can move the existing averageBlockTime calculation (using total block time) into the else part so to avoid double calculation

infrastructure/httpapi/handlers/validators.go
ysong42: Sure! Will change it!

infrastructure/httpapi/handlers/validators.go
ysong42: Good catch! Thanks for the advice! Will update this part!

infrastructure/httpapi/handlers/validators.go
```diff
@@ -285,14 +303,56 @@ func (handler *Validators) getAverageBlockTime() (*big.Float, error) {
...
-		new(big.Float).SetInt(totalBlockTime),
-		new(big.Float).SetInt64(int64(1000000)),
-	)
-	averageBlockTimeMilliSecond := new(big.Float).Quo(
-		totalBlockTimeMilliSecond,
-		new(big.Float).SetInt(totalBlockCount),
-	)
+	nRecentBlocks := big.NewInt(nRecentBlocksInInt)
+
+	isTotalBlockCountExcess := (totalBlockCount.Cmp(nRecentBlocks) == 1)
```
calvinaco, line 308: `isTotalBlockCountExcess` is probably not self-explanatory enough, what does it mean by excess? I will probably name it `hasNBlocksSinceGenesis`

infrastructure/httpapi/handlers/validators.go
```diff
@@ -285,14 +303,56 @@ func (handler *Validators) getAverageBlockTime() (*big.Float, error) {
...
+
+		// Case B
+		nRecentBlocksTotalTimeMilliSecond := new(big.Float).Quo(
+			new(big.Float).SetInt64(nRecentBlocksTotalTime),
+			new(big.Float).SetInt64(int64(1000000)),
+		)
+		averageBlockTimeMilliSecond = new(big.Float).Quo(
+			nRecentBlocksTotalTimeMilliSecond,
+			new(big.Float).SetInt(nRecentBlocks),
+		)
```
calvinaco, line 342:
```suggestion
		// Case B
		latestBlockHeight, err := handler.blockView.Count()
		if err != nil {
			return nil, fmt.Errorf("error fetching latest block height: %v", err)
		}
		latestBlockIdentity := block_view.BlockIdentity{MaybeHeight: &latestBlockHeight}
		latestBlock, err := handler.blockView.FindBy(&latestBlockIdentity)
		if err != nil {
			return nil, fmt.Errorf("error fetching latest block: %v", err)
		}

		// Find the nth block before latest block
		startBlockHeight := latestBlockHeight - nRecentBlocks.Int64()
		startBlockIdentity := block_view.BlockIdentity{MaybeHeight: &startBlockHeight}
		startBlock, err := handler.blockView.FindBy(&startBlockIdentity)
		if err != nil {
			return nil, fmt.Errorf("error fetching the start block: %v", err)
		}

		// Calculate total time in generating n recent blocks
		nRecentBlocksTotalTime := latestBlock.Time.UnixNano() - startBlock.Time.UnixNano()

		nRecentBlocksTotalTimeMilliSecond := new(big.Float).Quo(
			new(big.Float).SetInt64(nRecentBlocksTotalTime),
			new(big.Float).SetInt64(int64(1000000)),
		)
		averageBlockTimeMilliSecond = new(big.Float).Quo(
			nRecentBlocksTotalTimeMilliSecond,
			new(big.Float).SetInt(nRecentBlocks),
		)
```

calvinaco: Moving the comment case B to the beginning of the code block

calvinaco (commented):
Overall LGTM, left some suggestions below

infrastructure/httpapi/handlers/validators.go
ysong42: Agree! Will update the name.

infrastructure/httpapi/handlers/validators.go
ysong42: 👌

ysong42:
@calvinaco The code is updated!

calvinaco (approved):

PR #465, opened 2021-08-16
Problem: Simplify mechanism on parallelism

### LINKED ISSUE #467: Problem: Simplify mechanism on parallelism
Currently, parallelism of workers inside sync process is using custom channel to maintain. Suggest to use std `sync` package to simplify the mechanism.

davcrypto:
Solution: Fixed #467 by using std library `sync` to deal with parallelism


calvinaco:
Hello @davcrypto, can you create a corresponding issue and describe more in details inside and link it to this PR? Thank you.

usecase/syncstrategy/window.go
```diff
@@ -46,42 +45,33 @@ func (window *Window) Sync(
...
 	for height := beginHeight; height <= endHeight; height += 1 {
 		go func(height int64) {
 			commands, err := worker(height)
+			result := workResult{height, commands, nil}
+
 			if err != nil {
-				workResultCh <- workResult{height, nil, err}
+				logger.Errorf("received error from sync block worker #%d: %v", height, err)
+				result = workResult{height, nil, err}
 			}
```
calvinaco, line 63: I think the error is swallowed silently here.

usecase/syncstrategy/window.go
```diff
@@ -46,42 +45,33 @@ func (window *Window) Sync(
...
 
-			workResultCh <- workResult{height, commands, nil}
+			commandWindowMutex.Lock()
+			commandWindow.Put(result.height, result.commands)
+			commandWindowMutex.Unlock()
+
+			workersWaitGroup.Done()
```
calvinaco, line 68: One critque I will put is the use of Mutex here while in the original design it can be achieved using channel in a non-blocking manner to the goroutines. (but we probably should increase the channel size).

calvinaco: Also we probably need some error handling here and use `defer` to Unlock the window.

calvinaco: I still think the use of channel is better in terms of error handling and running things in parallel in general.

Although in this case since the critical section is at the end of goroutine so it may not hurt the correctness.

calvinaco (commented):
Thanks @davcrypto for the PR. I think it opens another solution but I personally have some critiques. Let's try to discuss it here before proceeding.

usecase/syncstrategy/window.go
davcrypto: I missed it, thanks for pointing out. Added error handler.

usecase/syncstrategy/window.go
davcrypto: Got your point. Actually, the original design was in a blocking manner already, we can see similar behaviour in here https://play.golang.org/p/BZArLAwhlym

As we can see the go routines is finally blocking each other since
```
for {
	temp := <- ch
	...
)
```
is the blocking statement

the same as original design
```
for {
	result := <-workResultCh
	...
)
```

davcrypto: Checked slice assignment is thread safe, we can improve from original design by removing the above mutex lock.

calvinaco: > Actually, the original design was in a blocking manner already, we can see similar behaviour in here

Yes, this is correct. This could be resolved by allocating a larger size channel.

Default channel: https://play.golang.org/p/oeTbXE0Ry4H
Explicitly sized channel: https://play.golang.org/p/io4EeKOYj4W

calvinaco: > Checked slice assignment (using index) is thread safe, we can improve from original design by removing the above mutex lock.

Sure, if that's the case then the Window is not "unsafe" anymore.

usecase/syncstrategy/window.go
```diff
@@ -46,42 +46,35 @@ func (window *Window) Sync(
...
 			if err != nil {
-				workResultCh <- workResult{height, nil, err}
+				logger.Errorf("received error from sync block worker #%d: %v", height, err)
+				return err
 			}
 
-			workResultCh <- workResult{height, commands, nil}
-		}(height)
-	}
+			result := workResult{height, commands, nil}
```
calvinaco, line 63: I think we no longer need the error to be stored here.

calvinaco (approved):
Awesome! The use of `errgroup` is an elegant way to handle error and can cancel context on first error occurs. utAck.

Since this will affect the core sync logic of the system, would be good to test it locally or staging environment with a longer time before deploying to production system.

usecase/syncstrategy/window.go
davcrypto: Updated, workers should be running in parallel

usecase/syncstrategy/window.go
davcrypto: Correct, removed the `workResult` struct

davcrypto:
@calvinaco Agree. For local, it is running for a while without issue. Lets test it in staging environment with a longer while

PR #470, opened 2021-08-16
Problem: pgmigrate.sh failed to install migrate under Golang v1.15 or below

### LINKED ISSUE #468: Problem: pgmigrate.sh failed to install migrate under Golang v1.15 or below
According to https://github.com/golang-migrate/migrate/tree/master/cmd/migrate#versioned, `golang-migrate` has different installation options based on Golang version

calvinaco:
Solution: (Fix #468) Update script to install migrate based on Golang version

### DIFF
--- a/pgmigrate.sh
+++ b/pgmigrate.sh
@@ -22,11 +22,74 @@ DB_DRIVER_URL="${DB_PROTOCOL}://${DB_USERNAME}:${DB_PASSWORD}@${DB_HOST}:${DB_PO
 
 echoerr() { echo "$@" 1>&2; }
 
+GOLANG_MIGRATE_VERSION="v4.14.1"
 check_migrate() {
     set +e
     command -v migrate > /dev/null
-    RET_VALUE=$?
+    if [[ $? != 0 ]]; then
+        RET_VALUE="NOT_FOUND"
+        set -e
+        return
+    fi
+    VERSION=$(migrate --version 2>&1)
+    if [[ "${VERSION}" != "${GOLANG_MIGRATE_VERSION}" ]]; then
+        if [[ $# == 1 && "$1" == "--verbose" ]]; then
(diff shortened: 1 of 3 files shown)


pgmigrate.sh
```diff
@@ -58,18 +121,27 @@ while [[ $# > 0 ]]; do
...
+    if [[ "${INSTALL_DEPENDENCY}" == 0 ]]; then
+        if [[ "${RET_VALUE}" == "NOT_FOUND" ]]; then
+            echoerr "Cannot find migrate. Provide \`--install-dependency\` to try to install it."
+            exit 1
+        fi
+    else
+        is_golang_1_16_and_above
+        if [[ "${RET_VALUE}" == 1 ]]; then
+#            install_golang_migrate_golang_1_16_and_above
+            install_golang_migrate_universal
```
davcrypto, line 135: Looks like no difference between `if` and `else` clause as install_golang_migrate_universal is running `install_golang_migrate_golang_1_15_and_below`

davcrypto: I guess `install_golang_migrate_golang_1_16_and_above` should be uncommented?

calvinaco: This is intended. I tried `install_golang_migrate_golang_1_16_and_above` at the beginning but I realize the `migrate` installed using this method has `migrate --version` always returns `dev` so it confuses the version check.

However, I want to keep the code here for future reference because this is the recommended way.

pgmigrate.sh
```diff
@@ -58,18 +121,27 @@ while [[ $# > 0 ]]; do
...
+        else
+            install_golang_migrate_golang_1_15_and_below
+        fi
```
calvinaco, line 138:
```suggestion
        install_golang_migrate_universal
```

pgmigrate.sh
davcrypto: I see

PR #480, opened 2021-08-20
Problem: pgmigrate script should not change directory after installing dependency

### LINKED ISSUE #479: Problem: pgmigrate script should not change directory after installing dependency

calvinaco:
Solution: (Fix #479) change to original directory after installation


ysong42 (approved):
LGTM!

PR #485, opened 2021-08-24
Problem: MsgRecvPacket success value is inverted

### LINKED ISSUE #466: Problem: MsgRecvPacket success value is inverted
Because of a previous bug on Cosmos SDK, the `success` value is inverted on purpose. As it is fixed now, we need to invert it again.

To extend from this task, if there's a way for us to tell which Cosmos SDK the chain is running, then we can include different versions' parsing logic in the code. This could potentially be done by querying the `/node_info` endpoint.

https://github.com/crypto-com/chain-indexing/blob/287a6cf023a118ad830f456cdb7063dfe1f87063/usecase/parser/ibcmsg/ibcmsg.go#L773

calvinaco:
Solution: (Fix #466) Parse message success value according to Cosmos SDK version

### DIFF
--- a/cmd/chain-indexing/app.go
+++ b/cmd/chain-indexing/app.go
@@ -129,8 +129,9 @@ func CliApp(args []string) error {
 				cliConfig.DatabasePort = primptr.Int32(int32(ctx.Int("dbPort")))
 			}
 
-			config := Config{
-				fileConfig,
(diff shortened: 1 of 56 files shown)


calvinaco:
Closing it as we will be managing different versions support using branch instead of dynamic config.

PR #492, opened 2021-08-25
Problem: Missing support to SoloMachine IBC transfer and json.Number will marshal to number but not string

### LINKED ISSUE #487: Problem: json.Number will marshal to number but not string
`json.Number` has the following limitations:
1. It does not support `uint64`
2. It always marshal to number, which may creates problem when being parsed in some langauges

For better handling of number across different languages, add a new json struct to support
- Unmarshal from both string and numeric representation of the underlying number
- Marshal to number and string based on the range of the number

### LINKED ISSUE #488: Problem: Missing support to SoloMachine IBC transfer
The amount inside packet data is in number but not string

calvinaco:
Problem: Missing support to SoloMachine IBC transfer
Problem: json.Number will marshal to number but not string

Solution: (Fix #487) (Fix #488) Add custom json number types and support SoloMachine IBC transfer

### DIFF
--- a/usecase/model/ibc/ibc.go
+++ b/usecase/model/ibc/ibc.go
@@ -1,5 +1,7 @@
 package ibc
 
+import "github.com/crypto-com/chain-indexing/internal/json"
+
 type Height struct {
 	RevisionNumber uint64 `mapstructure:"revision_number" json:"revisionNumber,string"`
 	RevisionHeight uint64 `mapstructure:"revision_height" json:"revisionHeight,string"`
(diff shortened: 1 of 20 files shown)


usecase/model/ibc/ibc.go
```diff
@@ -45,8 +47,8 @@ type Packet struct {
...
 
 type FungibleTokenPacketData struct {
-	Sender   string `mapstructure:"sender" json:"sender"`
-	Receiver string `mapstructure:"receiver" json:"receiver"`
-	Denom    string `mapstructure:"denom" json:"denom"`
-	Amount   uint64 `mapstructure:"amount" json:"amount,string"`
+	Sender   string      `mapstructure:"sender" json:"sender"`
+	Receiver string      `mapstructure:"receiver" json:"receiver"`
+	Denom    string      `mapstructure:"denom" json:"denom"`
+	Amount   json.Uint64 `mapstructure:"amount" json:"amount"`
```
devashishdxt, line 53: The wire format actually contains `amount` as  `string`. I hope this'll work with `string`?

davcrypto (approved):

usecase/model/ibc/ibc.go
davcrypto: I think so @devashishdxt, tested a bit on my local and its green. `json.Uint64` should aims to deal with both `uint64` and `string` format

PR #494, opened 2021-08-25
ADD missing solo machine message

### LINKED ISSUE #493: Problem: Missing some message type in solo machine transactions
Missing `MsgCreateClient` and `MsgRecvPacket` in solo machine transactions

davcrypto:
Solution: ( fix #493 ) add back logic for validations

### DIFF
--- a/usecase/parser/ibcmsg/ibcmsg.go
+++ b/usecase/parser/ibcmsg/ibcmsg.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/base64"
 	"fmt"
+	"github.com/crypto-com/chain-indexing/internal/slice"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


internal/json/uint32.go
```diff
@@ -0,0 +1,58 @@
...
+func (u Uint32) Uint32() uint32 {
+	return u.v
+}
+
+func (u Uint32) String() string {
+	return strconv.Itoa(int(u.v))
+}
+
+func (u Uint32) MarshalJSON() ([]byte, error) {
+	// uint32 number can be handled by all languages as number
```
calvinaco, line 26:
```suggestion
```

usecase/parser/ibcmsg/ibcmsg.go
```diff
@@ -21,15 +22,20 @@ import (
...
 	msgCommonParams event.MsgCommonParams,
 	txsResult model.BlockResultsTxsResult,
 	msgIndex int,
 	msg map[string]interface{},
 ) []command.Command {
 	clientStateType := msg["client_state"].(map[string]interface{})["@type"]
-	if clientStateType != "/ibc.lightclients.tendermint.v1.ClientState" {
-		// TODO: SoloMachine and Localhost LightClient
+
+	if c, ok := clientStateType.(string); ok && !slice.ContainString(allowedClientStateTypes, c) {
```
ysong42, line 38: I think if we assume the `clientStateType` is always a string, then we could replace `ok` with `_`.

What do you think?

PR #511, opened 2021-09-02
Problem: missing IBC channel debug projection

### LINKED ISSUE #509: Problem: missing IBC channel debug projection

ysong42:
Solution: fixed #509 

### How to enable?

In config file, use `IBCChannelTxMsgTrace` projection instead of `IBCChannel` projection.

### Added a new table `view_ibc_channel_message`

Table schema:
```
CREATE TABLE view_ibc_channel_messages (
    id BIGSERIAL,
    channel_id VARCHAR NOT NULL,
    block_height BIGINT NOT NULL,
    source_channel VARCHAR NOT NULL,
    destination_channel VARCHAR NOT NULL,
    denom VARCHAR NOT NULL,
    amount VARCHAR NOT NULL,
    success VARCHAR NOT NULL,
    error VARCHAR NOT NULL,
    message_type VARCHAR NOT NULL,
    message JSONB NOT NULL,
    updated_bonded_tokens JSONB NOT NULL,
    PRIMARY KEY(id)
);
```

### DIFF
--- a/projection/ibc_channel/ibc_channel.go
+++ b/projection/ibc_channel/ibc_channel.go
@@ -32,8 +32,12 @@ type Config struct {
 }
 
 func NewIBCChannel(logger applogger.Logger, rdbConn rdb.Conn, config Config) *IBCChannel {
+	projectionID := "IBCChannel"
+	if config.EnableTxMsgTrace {
+		projectionID = "IBCChannelTxMsgTrace"
+	}
 	return &IBCChannel{
-		rdbprojectionbase.NewRDbBase(rdbConn.ToHandle(), "IBCChannel"),
+		rdbprojectionbase.NewRDbBase(rdbConn.ToHandle(), projectionID),
 
 		rdbConn,
 		logger,
--- a/projection/ibc_channel/ibc_channel.go
+++ b/projection/ibc_channel/ibc_channel.go
@@ -2,6 +2,7 @@ package ibc_channel
 
 import (
 	"fmt"
+	"strconv"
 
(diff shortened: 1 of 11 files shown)


projection/ibc_channel/ibc_channel.go
```diff
@@ -21,14 +23,22 @@ type IBCChannel struct {
 
 	rdbConn rdb.Conn
 	logger  applogger.Logger
+
+	config Config
+}
+
+type Config struct {
+	EnableTxMsgTrace bool
```
davcrypto, line 31: Any reason we need a flag for the trace?

ysong42: I am trying to follow the way how `NFT` projection in debug mode 😂 . Maybe you could have a look on it.

The trace is not always enabled, so when using projection `IBCChannel`, it is not enabled.
Only when using projection `IBCChannelTxMsgTrace`, the trace will be enabled, which will write those msgs to the newly introduced DB table.

davcrypto: Does it mean the trace is for local use only? since once the flag is off for a while, it will lose the ability to provide the full trace unless someone reindex the whole ibc_channel

ysong42: Yes, I think so. I think mainly for debugging on local and testnet. But need to confirm with Calvin. I remember he mentioned that some people might not want us to trace this on mainnet?

calvinaco (approved):
LGTM!

PR #515, opened 2021-09-06
Problem: missing created_at_block_time and last_activity_block_time sorting params for ibc channel list API

### LINKED ISSUE #514: Problem: missing `created_at_block_time` and `last_activity_block_time` sorting params for ibc channel list API

ysong42:
Solution: fixed #514 

Provide optional query params for `ibc/channels`.

```
# sort by fields
order= lastActivityBlockHeight.asc
order= lastActivityBlockHeight.desc ------> default sorting
order= createdAtBlockTime.asc
order= createdAtBlockTime.desc

# optional filter by status
status=true
status=false
# if `status` filter NOT provided, will return channel in both status
```

### DIFF
--- a/infrastructure/httpapi/handlers/ibc_channel.go
+++ b/infrastructure/httpapi/handlers/ibc_channel.go
@@ -43,13 +43,13 @@ func (handler *IBCChannel) ListChannels(ctx *fasthttp.RequestCtx) {
 
 	queryArgs := ctx.QueryArgs()
 	if queryArgs.Has("order") {
-		if string(queryArgs.Peek("order")) == "created_at_block_time.desc" {
+		if string(queryArgs.Peek("order")) == "createdAtBlockTime.desc" {
 			listOrder.MaybeCreatedAtBlockTime = &view_order_desc
-		} else if string(queryArgs.Peek("order")) == "created_at_block_time.asc" {
+		} else if string(queryArgs.Peek("order")) == "createdAtBlockTime.asc" {
 			listOrder.MaybeCreatedAtBlockTime = &view_order_asc
-		} else if string(queryArgs.Peek("order")) == "last_activity_block_time.desc" {
+		} else if string(queryArgs.Peek("order")) == "lastActivityBlockTime.desc" {
 			listOrder.MaybeLastActivityBlockTime = &view_order_desc
-		} else if string(queryArgs.Peek("order")) == "last_activity_block_time.asc" {
+		} else if string(queryArgs.Peek("order")) == "lastActivityBlockTime.asc" {
 			listOrder.MaybeLastActivityBlockTime = &view_order_asc
 		}
 	}
--- a/infrastructure/httpapi/handlers/ibc_channel.go
+++ b/infrastructure/httpapi/handlers/ibc_channel.go
@@ -37,17 +37,24 @@ func (handler *IBCChannel) ListChannels(ctx *fasthttp.RequestCtx) {
 		return
 	}
 
-	channelIdOrder := view.ORDER_ASC
+	var listOrder ibc_channel_view.IBCChannelsListOrder
+	view_order_desc := view.ORDER_DESC
+	view_order_asc := view.ORDER_ASC
+
 	queryArgs := ctx.QueryArgs()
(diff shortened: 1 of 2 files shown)


infrastructure/httpapi/handlers/ibc_channel.go
```diff
@@ -37,17 +37,24 @@ func (handler *IBCChannel) ListChannels(ctx *fasthttp.RequestCtx) {
 		return
 	}
 
-	channelIdOrder := view.ORDER_ASC
+	var listOrder ibc_channel_view.IBCChannelsListOrder
+	view_order_desc := view.ORDER_DESC
+	view_order_asc := view.ORDER_ASC
```
calvinaco, line 42: Why copying the value here?

projection/ibc_channel/view/ibc_channels.go
```diff
@@ -450,12 +451,26 @@ func (ibcChannelsView *IBCChannels) List(order IBCChannelsListOrder, pagination
 		"bonded_tokens",
 	).From(
 		"view_ibc_channels",
+	).Where(
+		"status = ?", "true", // Only if status is true, the channel is in `opened` status
```
calvinaco, line 455: Shall we make this a filter option?

calvinaco (commented):
Left some comment above

infrastructure/httpapi/handlers/ibc_channel.go
ysong42: Because `view.ORDER_DESC` and `view.ORDER_ASC` are constants, they are not addressable.

Here is a reference: https://stackoverflow.com/questions/35146286/find-address-of-constant-in-go

projection/ibc_channel/view/ibc_channels.go
ysong42: Sure! Let me make it enable by default.

Some extra context here. 

At the moment, as we did not handle ChannelClose and ChannelCloseOnTimeout (not sure if I write their name correctly here 😂 ), `status` field only indicated that the channel is successfully established, that is when `OpenChannelAck` or `OpenChannelConfirm` is detected, `status` will be set as `true`.

In the future, when we handle channel close properly, this field could really represent that the channel is `closed`.

ysong42:
@calvinaco @davcrypto Thanks for the review! The PR is updated. Need help to have a second round review 🙏 

I added a new optional filter `status`.
- `status=true`, only filtered channel with true status
- `status=false`, only filtered channel with false status
- when NOT provided, returned channel in both status

One more thing would like to mention is that for `bool` pointer, we also need to create a variable to store it first then could get its pointer. 

a reference: https://stackoverflow.com/questions/28817992/how-to-set-bool-pointer-to-true-in-struct-literal

infrastructure/httpapi/handlers/ibc_channel.go
calvinaco: Try `primptr`?

infrastructure/httpapi/handlers/ibc_channel.go
```diff
@@ -37,17 +37,36 @@ func (handler *IBCChannel) ListChannels(ctx *fasthttp.RequestCtx) {
...
+		}
+	}
+
+	var listFilter ibc_channel_view.IBCChannelsListFilter
+	view_filter_status_true := true
+	view_filter_status_false := false
+
+	if queryArgs.Has("status") {
+		if string(queryArgs.Peek("status")) == "true" {
+			listFilter.MaybeStatus = &view_filter_status_true
```
calvinaco, line 63: Similarly try `primptr` here

calvinaco (approved):
LGTM, left some enhancement suggestions.

infrastructure/httpapi/handlers/ibc_channel.go
ysong42: Wow, that's handy! Let me change it!

infrastructure/httpapi/handlers/ibc_channel.go
ysong42: 👍

PR #526, opened 2021-09-24
Problem: missing logic to parse MsgExec internal messages

### LINKED ISSUE #525: Problem: missing logic to parse MsgExec internal messages

ysong42:
Solution: fixed #525

### DIFF
--- a/usecase/parser/msg.go
+++ b/usecase/parser/msg.go
@@ -137,6 +137,7 @@ func ParseBlockResultsTxsMsgToCommands(
 					MsgCommonParams: msgCommonParams,
 					Msg:             msg,
 					MsgIndex:        msgIndex,
+					ParserManager:   parserManager,
 				})
 			}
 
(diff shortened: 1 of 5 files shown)


usecase/parser/msg.go
```diff
@@ -1294,22 +1295,67 @@ func ParseMsgExec(
...
+		parser := parserParams.ParserManager.GetParser(utils.CosmosParserKey(innerMsgType), utils.ParserBlockHeight(blockHeight))
+
+		msgCommands := parser(utils.CosmosParserParams{
+			AddressPrefix:   parserParams.AddressPrefix,
+			StakingDenom:    parserParams.StakingDenom,
+			TxsResult:       parserParams.TxsResult,
+			MsgCommonParams: parserParams.MsgCommonParams,
+			Msg:             innerMsg,
+			MsgIndex:        parserParams.MsgIndex,
+			ParserManager:   parserParams.ParserManager,
```
davcrypto, line 1352: why we need to pass `ParserManager` here?

davcrypto (commented):
Thanks @ysong42. I left a comment for discussion. Other than that LGTM.

usecase/parser/msg.go
ysong42: This is mainly for parsing inner msgs for MsgExec. After parsing MsgExec, we will know its inner msg type, and then will need this `ParserManager` to retrieve the corresponding parser for parsing the inner msg.

davcrypto: would the inner messages need `ParserManager`?

usecase/parser/msg_exec_test.go
```diff
@@ -78,6 +97,22 @@ var _ = Describe("ParseMsgCommands", func() {
...
+
+			Expect(json.MustMarshalToString(createMsgSendEvent)).To(Equal(
+				strings.Replace(
+					regex.ReplaceAllString(expectedInnerMsg, ""),
+					"{UUID}",
+					createMsgSendEvent.UUID(),
+					-1,
+				),
+			))
+
```
calvinaco, line 115: Can we have 1-2 more examples of MsgExec with a more complicated inner message? Want to make sure the parser is correct for diff message types.

ysong42: Sure! Let me check if I could find more MsgExec related tx.

usecase/parser/msg.go
ysong42: @davcrypto I am so sorry... I missed the message 🙏  I will always refresh the page later on...

Yes, the inner messages will need the `ParserManager`. As the `ParseMsgExec()` could only parsed out a `model.RawMsgExec`, which does not handle the inner message.

We need to find the inner message type first, and then get the parser function from `ParserManager`.

davcrypto (approved):

usecase/parser/msg_exec_test.go
ysong42: @calvinaco Hi, I just added one more MsgExec example, whose inner msg is `MsgDelegate`.

PR #530, opened 2021-10-06
ibc_channel projection unit test

davcrypto:


ysong42 (commented):
Super cool! Looks good to me!

I think the only issue is, the new test cases are more `unit test` like. While previous test cases in other projections seem to be more `integration` like.

We might need some input from @calvinaco here 😂

ysong42 (approved):

projection/ibc_channel/ibc_channel_test.go
```diff
@@ -0,0 +1,828 @@
...
+							TendermintLightClientConsensusState: ibc_model.TendermintLightClientConsensusState{},
+						},
+						ClientID: "ClientID",
+					},
+				},
+			},
+			MockFunc: func(mockCoon *test.MockRDbConn) (mocks []*testify_mock.Mock) {
+				mockTx := NewMockRDbTx()
+				mocks = append(mocks, &mockTx.Mock)
+				mockCoon.On("Begin").Return(mockTx, nil)
```
calvinaco, line 91:
```suggestion
				mockConn.On("Begin").Return(mockTx, nil)
```

projection/ibc_channel/ibc_channel_test.go
```diff
@@ -0,0 +1,828 @@
...
+		mockRDbConn := NewMockRDbConn()
+		mocks := tc.MockFunc(mockRDbConn)
+		mocks = append(mocks, &mockRDbConn.Mock)
+
+		projection := NewIBCChannelProjection(mockRDbConn)
+		err := projection.HandleEvents(1, tc.Events)
+		assert.NoError(t, err)
+
+		for _, m := range mocks {
+			m.AssertExpectations(t)
```
calvinaco, line 801: May I know what will be assert upon running this line? Is that all `mockTx.On(...)` is expected to be run at least once?

projection/ibc_channel/ibc_channel_test.go
```diff
@@ -0,0 +1,828 @@
...
+	entity_event "github.com/crypto-com/chain-indexing/entity/event"
+	"github.com/crypto-com/chain-indexing/infrastructure/pg"
+	"github.com/crypto-com/chain-indexing/internal/json"
+
+	//"github.com/crypto-com/chain-indexing/internal/json"
+	"github.com/crypto-com/chain-indexing/projection/ibc_channel"
+	event_usecase "github.com/crypto-com/chain-indexing/usecase/event"
+	ibc_model "github.com/crypto-com/chain-indexing/usecase/model/ibc"
+	"github.com/stretchr/testify/assert"
+	testify_mock "github.com/stretchr/testify/mock"
```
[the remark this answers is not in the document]
davcrypto: got it, let me organise the imports

projection/ibc_channel/ibc_channel_test.go
calvinaco: Got it, one critics of mocking the RDb would be that the test will be very dependant on the implementation but what we concern about is more on the outcome.

Let's say one day we have optimized the handling logic, or do single insert all instead of multiple insertions, this will break the test and is hard to update.

calvinaco: Overall I like the framework and I also saw it in many other Golang projects which I think is clear. But what do assert I think we can think about it.

davcrypto: I see your concern. There is a `Repeatability` for each call (if same method name and same arguments). Test cases can use `.Once()`, `.Tiwce()` or` Times(int)` to indicated it. By default, the assertion would already pass if no `Repeatability` was specified.

ysong42 (approved):
Thanks! LGTM 👍

PR #542, opened 2021-10-18
Problem: sample config files are not up-to-dated

calvinaco:


ysong42 (approved):

ysong42:
I have a question on the config 🙏 

Why do we enable `BridgePendingActivity` projection for `croeseid-4` network? 
In my understanding, it seems that `BridgePendingActivity` and cronjob `BridgeActivityMatcher` need to work together.

Thanks!

PR #574, opened 2021-10-30
Problem: Bridge Activities API missing id and time filter

### LINKED ISSUE #573: Problem: Bridge Activities API missing id and time filter

calvinaco:
Solution: (Fix #573) Add greater than filter

### DIFF
--- a/infrastructure/httpapi/handlers/bridges.go
+++ b/infrastructure/httpapi/handlers/bridges.go
@@ -178,14 +178,32 @@ func (handler *Bridges) ListActivitiesByNetwork(ctx *fasthttp.RequestCtx) {
 		}
 	}
 
-	addressFilter := bridge_activitiy_view.BridgeActivitiesListAddressFilter{
-		MaybeCronosAddress:         nil,
-		MaybeCryptoOrgChainAddress: nil,
-	}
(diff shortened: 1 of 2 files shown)


infrastructure/httpapi/handlers/bridges.go
```diff
@@ -154,11 +179,34 @@ func (handler *Bridges) ListActivitiesByNetwork(ctx *fasthttp.RequestCtx) {
 	}
 
 	filter := bridge_activitiy_view.BridgeActivitiesListFilter{
+		MaybeIdGt:        nil,
+		MaybeCreatedAtGt: nil,
+		MaybeUpdatedAtGt: nil,
+	}
+	if queryArgs.Has("id.gt") {
```
davcrypto, line 186: in this way, `id.gt`, `createdAt.gt` and `updatedAt.gt` will be maturely exclusion. Do we intended to do?

calvinaco: Yes this is expected behaviour. It is intended for downstream to scan through the records. But I think there's no harm for me to have them work in parallel. Let me update it.

davcrypto (approved):

PR #593, opened 2021-11-08
Problem: Validator projection catching up speed is slow

### LINKED ISSUE #586: Problem: Validator projection catching up speed is slow
After 2 days running with a clean DB, the `Validator` projection only catch up around 400k block, while in total we have around 3000k block on mainnet PROD.

It will take around 15 days to do re-index for this projection.

Screenshot 2021-11-03 at 11 21 31 AM

We need to investigate what is causing this bottleneck and fix it.

ysong42:
Solution: fixed #586 

The main bottleneck is in handling `BlockCreated` event, we have some code that is updating all active validators' uptime. Previously, we are doing it in a loop manner, each validator will do one update SQL.

In my improved version, there are mainly two changes:
- update validators up time in batch mode, merge many update SQL into one
- Only updated uptime related columns, not updating all columns of validator

Update multiple rows reference: https://stackoverflow.com/a/26715934

## Benchmark

**the improvement is at `BlockCreated-updateValidator` column**

Table columns:
image

Before optimization:
image

After optimization:
image

Benchmark data files (local indexer running against Crypto.org mainnet):
[before-optimize-validator.csv](https://github.com/crypto-com/chain-indexing/files/7494109/before-optimize-validator.csv)
[after-optimize-validator.csv](https://github.com/crypto-com/chain-indexing/files/7494114/after-optimize-validator.csv)


ysong42:
I am not using squirrel library to generate the SQL. As update with from statement is not supported by the library yet...

https://github.com/Masterminds/squirrel/issues/82
https://github.com/Masterminds/squirrel/issues/291

davcrypto (approved):

davcrypto:
LGTM

PR #596, opened 2021-11-09
Update proposal list sorting

davcrypto:
Update proposal list to sort by numeric id


calvinaco (approved):
LGTM. Do you prefer to have this merge first or try to update the column designs in the proposal table?

ysong42 (approved):

ysong42:
I think in the next step, maybe we shall consider to merge the two columns into one?

Maybe we could assume proposal_id is always a integer and do a conversion, then write it to DB?

davcrypto:
I suggest to keep to `proposal_id` first to be consistent with cosmos first. And for sorting, will using the `id`

PR #598, opened 2021-11-09
Problem: Current logic of "failed IBC transfer" maybe inaccurate

### LINKED ISSUE #590: Problem: Current logic of "failed IBC transfer" maybe inaccurate
Right now the IBC success count and rate is treating timeout transfer as failure. However, these timeout transfers is more of user error than relayer error. In fact in these case relayer is doing its job to relay these timeout failure correctly.

The idea of success rate should be re-defined to reflect the failure of relayer(s), but not packet timeout.

ysong42:
Solution: fixed #590 

## Changes:
- Renamed ibc_channel table columns `total_transfer_*` to `total_relay_*`.
- Success rate now indicates the ratio of successfully relayed `IBC Transfer`
- Updated related test cases

## A small reminder:
If you want to test newly added migration script, you will need to update `MIGRATION_GITHUB_TARGET`.

First, you need to pushed your code to a branch, then change the const:
`MIGRATION_GITHUB_TARGET = "github://%s:%s@crypto-com/chain-indexing/%s#<YOUR_BRANCH_NAME>"`

For example, to test this branch locally, you will need to change it to:
`MIGRATION_GITHUB_TARGET = "github://%s:%s@crypto-com/chain-indexing/%s#fix/590-relay-out-success-rate"`

### DIFF
--- a/projection/ibc_channel/ibc_channel.go
+++ b/projection/ibc_channel/ibc_channel.go
@@ -268,27 +268,27 @@ func (projection *IBCChannel) HandleEvents(height int64, events []event_entity.E
 			}
 
 			channel := &ibc_channel_view.IBCChannelRow{
-				ChannelID:                    msgIBCChannelOpenInit.Params.ChannelID,
-				PortID:                       msgIBCChannelOpenInit.Params.PortID,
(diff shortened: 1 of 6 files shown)


calvinaco (approved):
LGTM!

PR #607, opened 2021-11-10
Problem: Bridge Activities fail status should be FailedOnChain but not Failed

### LINKED ISSUE #604: Problem: Bridge Activities fail status should be `FailedOnChain` but not `Failed`

calvinaco:
Solution: (Fix #604) Change fail status to FailedOnChain


ysong42 (approved):
LGTM!

One small reminder! 

When we introduce new migration scripts, will need to create a new tag and change the `config.github_api.migration_repo_ref` to match the new tag. So that `chain-indexing-server` will fetch the correct migrations file from `chain-indexing`. 🙏

PR #615, opened 2021-11-17
Problem: MsgTimeout could be relayed multiple times

### LINKED ISSUE #608: Problem: MsgTimeout could be relayed multiple times
## How do I find this?

image

`Ralay out success rate` is more than 1... And it is weird that `total success` is larger than `total`...

Then I check the staging DB and find:

image

Next, go to explorer and check the related txs on explorer, I find they are using same `sequence`.

https://chain.3ona.co/explorer/croeseid4/block/1303950
https://chain.3ona.co/explorer/croeseid4/block/1303953
https://chain.3ona.co/explorer/croeseid4/block/1303954

In a word, one `MsgTransfer` (IBC Transfer) triggered two `MsgTimeout`s (IBC Packet Timeout).

## How to fix this?

I checked `ibc-go` implementation, and find the `MsgTimeout` will still show up in Block even it is relayed for the second time.

This is a similar case that we encountered before, the msg is relayed twice, and we should ignore the second one.

For the second `MsgTimeout`, the chain will do `no-op`, the refund will not occur for a second time. But the chain will still broadcast `timeout_packet` event.

Code reference:
https://github.com/cosmos/ibc-go/blob/823ef675bbf38a48faa3ba670fb6b8228d58b9fd/modules/core/keeper/msg_server.go#L552
https://github.com/cosmos/ibc-go/blob/823ef675bbf38a48faa3ba670fb6b8228d58b9fd/modules/core/04-channel/keeper/events.go#L119
https://github.com/cosmos/ibc-go/blob/823ef675bbf38a48faa3ba670fb6b8228d58b9fd/modules/core/04-channel/keeper/timeout.go#L77

Sample Block Results:
https://testnet-croeseid-4.crypto.org:26657/block_results?height=1303953 (log contains a `timeout` event)
https://testnet-croeseid-4.crypto.org:26657/block_results?height=1303954 (log not contain a `timeout` event)

I will suggest the following changes: 
- in parser function, check if `timeout` event exist in `log`
- introduce a `NoOperation` boolean field for `MsgTimeout`. And maybe also introduce this to `MsgTimeoutOnClose`, `MsgRecvPacket` and `MsgAcknowledgement`?

davcrypto:
Solution: Fixed #608 by adding AlreadyRelayed type messages


calvinaco (commented):
@davcrypto I notice the parsing logic is unchanged, does it mean that the `AlreadyRealyed...` messages are still having `success:false`?

usecase/parser/ibcmsg/ibcmsg.go
```diff
@@ -709,6 +709,8 @@ func ParseMsgRecvPacket(
 		// https://github.com/cosmos/ibc-go/blob/760d15a3a55397678abe311b7f65203b2e8437d6/modules/core/04-channel/keeper/packet.go#L239
 		// https://github.com/cosmos/ibc-go/blob/760d15a3a55397678abe311b7f65203b2e8437d6/modules/core/keeper/msg_server.go#L508
 
+		// TODO: create MsgAlreadyRelayedRecvPacket
+
```
ysong42, line 713:
```suggestion
```

ysong42 (dismissed):
LGTM!

ysong42:
Thanks Calvin for catching that. 

Yes, I think `MsgAlreadyRelayedXXX` should not have `success` field.

Maybe we could use the `block_results.log` in the test cases, to compare the different events between `MsgXXX` and `MsgAlreadyRelayed`. Then define more specific structs for the new Msgs.

Thanks!

davcrypto:
@calvinaco @ysong42 Yea, the fact that i haven't change the success field is that to maintain the same struct to the original message. Otherwise, we will need to maintain two set of struct separately for each message. Or you guys still though that a separated struct make more sense?

ysong42:
Yes, I think it is better to create another set of `Params` struct.

The format could be  simple, take `MsgAlreadyRelayedRecvPacketParams` for example, I will suggest the below:

```
type MsgRecvPacketParams struct {
	RawMsgRecvPacket

	Application                  string                                `json:"application"`
	MessageType                  string                                `json:"messageType"`
}
```

Thanks!

davcrypto:
Ok, if then i would remove the `MaybeFungibleTokenPacketData` assignment only for better maintainability @ysong42 since `MsgAlreadyRelayedXXX` better have the same struct when corresponding `MsgXXX`

davcrypto:
Updated

ysong42 (approved):
LGTM!

calvinaco:
@davcrypto can you elaborate on 
> Ok, if then i would remove the MaybeFungibleTokenPacketData assignment only for better maintainability @ysong42 since MsgAlreadyRelayedXXX better have the same struct with corresponding MsgXXX

If we do it this way, projection developer won't gain much benefits because if the follow the Msg types they will see the `MaybeFungibleTokenPacketData` is there and there's a success field inside.

PR #617, opened 2021-11-19
Problem: MsgAcknowledgement could be relayed multiple times

davcrypto:
Solution: Separated struct for already relayed messages


ysong42 (approved):
Thanks for the PR! LGTM!

PR #619, opened 2021-11-23
Refactor init and run migration in projection.OnInit()

ysong42:


infrastructure/pg/migrate.go
```diff
@@ -118,3 +118,74 @@ func (m *Migrate) Reset() error {
 
 	return nil
 }
+
+// Get the migration source files from Github, init the migrate and then run it.
+func InitAndRunMigrateFromGithub(
```
davcrypto, line 123: Too much param, the caller side should be able to construct migrationSourceURL itself

ysong42: How about this one?
```
func (projection *Account) OnInit() error {
	migrationSourceURL := pg.GithubMigrationSourceURL(
		appprojection.MIGRATION_GITHUB_TARGET,
		projection.config.GithubAPIUser,
		projection.config.GithubAPIToken,
		MIGRATION_DIRECOTRY,
		projection.config.MigrationRepoRef,
	)
	databaseURL := pg.MigrationDBConnString(
		projection.rdbConn.(*pg.PgxConn).ConnString(),
		MIGRATION_TABLE_NAME,
	)
	if err := pg.InitAndRunMigrate(migrationSourceURL, databaseURL); err != nil {
		projection.logger.Errorf("failed to init and run migration: %v", err)
		return err
	}

	return nil
}
```

davcrypto: Good and change `InitAndRunMigrate ` to `migrate`/`migrateFromGitHub`

ysong42: Sure, np~

davcrypto (approved):

calvinaco (approved):
The refactoring looks good to me. But I am thinking if we can take it further.

In particular the GitHub credential is not suppose to be changed across different projections, and the folder could possibly be simply using the projection name.

I imagined we could make the GitHub migration helper a singleton, and we can introduce a new projection base struct say RdbProjectionWithMigtatuon that always do migration based on the above rule.

On the other hand, the GitHub migration singleton can be further refactor to fulfil an interface, and we can introduce a file based migration system. Then GitHub will no longer be a dependency and give end user extra choices.

ysong42:
> The refactoring looks good to me. But I am thinking if we can take it further.
> 
> In particular the GitHub credential is not suppose to be changed across different projections, at least for our built-in projection and the folder could possibly be simply using the projection name.
> 
> I imagined we could keep an instance of the GitHub migration helper, and we can introduce a new projection base struct say RdbProjectionWithMigtatuon that leverage the instance always do migration based on the above rule.
> 
> On the other hand, the GitHub migration helper can be further refactor to fulfil an interface, and we can introduce a file based migration system. Then GitHub will no longer be a dependency and give end user extra choices.

Thanks for the suggestion. I will try to implement it today!

infrastructure/pg/migrationhelper/migrationhelper.go
```diff
@@ -0,0 +1,17 @@
+package migrationhelper
+
+type MigrationHelper interface {
+	InitAndRunMigrate()
```
davcrypto, line 4: `InitAndRunMigrate ` -> `migrate`

infrastructure/pg/migrationhelper/github/github.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/golang-migrate/migrate/v4"
+
+	appprojection "github.com/crypto-com/chain-indexing/projection"
+)
+
+type GithubMigrationHelper struct {
+	Config Config
+
+	MaybeSourceURL   *string
```
davcrypto, line 15: No need for pointer, we could use empty value when doing validation

infrastructure/pg/migrationhelper/github/github.go
```diff
@@ -0,0 +1,74 @@
...
+	MaybeDatabaseURL *string
```
davcrypto, line 16: same as above

usecase/projection/base.go
```diff
@@ -68,3 +98,31 @@ func (base *Base) Id() string {
...
 }
+
+func setupDefaultGithubMigration(
+	projectionId string,
+	migrationHelper *github_mh.GithubMigrationHelper,
+) {
+	projectionSnakeId := strcase.ToSnake(projectionId)
+
+	if migrationHelper.MaybeSourceURL == nil {
+		migrationDirectory := fmt.Sprintf(github_mh.MIGRATION_DIRECTORY_FORMAT, projectionSnakeId)
```
davcrypto, line 109: Can we just panic if SourceURL is not valid? Looks like there is no generic way to build a path for every projection/future entity

usecase/projection/base.go
```diff
@@ -68,3 +98,31 @@ func (base *Base) Id() string {
 func (base *Base) Config() interface{} {
 	return base.config
 }
+
+func setupDefaultGithubMigration(
+	projectionId string,
```
davcrypto, line 103: `projectionId` -> `id`

projection/bridge_activity/bridge_pending_activity/bridge_pending_activity.go
```diff
@@ -41,17 +40,44 @@ type Config struct {
...
+				MaybeMigrationHelper: github_mh.NewGithubMigrationHelper(
+					github_mh.Config{
+						Config:     *config,
+						ConnString: rdbConn.(*pg.PgxConn).ConnString(),
+					},
+					migrationSourceURL,
+					"",
+				),
+			},
+		),
```
calvinaco, line 80: Can we pass the migration helper as an argument too? That way user can decide to use non-GitHub way to instantiate this projection.

ysong42: Oh, I also introduced a `FilesystemMigrationHelper`. The MaybeMigrationHelper's type is `MigrationHelper` interface.

User could switch another type of migration helper.

usecase/projection/base.go
ysong42: Will keep `projectionId` as discussed.

usecase/projection/base.go
ysong42: Will move the execute `MigrationHelper.Migrate()` logic to projection's `OnInit()`.

projection/bridge_activity/bridge_pending_activity/bridge_pending_activity.go
davcrypto: I think we can use migration helper as an projection object argument and use it in projection.OnInit().

ysong42: oh, I get it now. I will put the `XXXMigrationHelper` initialization logic on the caller side. Will change it accordingly!

ysong42:
@calvinaco @davcrypto The code is updated again!

Two major changes:
- Extract `NewXXXMigrationHelper` logic to caller side. 
- Move `Migrate()` execution to `projection.OnInit()`.

Sample code on `chain-indexing-server` side:

image

infrastructure/pg/migrationhelper/filesystem/filesystem.go
```diff
@@ -0,0 +1,45 @@
...
+func NewFilesystemMigrationHelper(sourceURL, databaeURL string) *FilesystemMigrationHelper {
+	return &FilesystemMigrationHelper{
+		SourceURL:   sourceURL,
+		DatabaseURL: databaeURL,
+	}
+}
+
+// Implement MigrationHelper interface
+func (fmh *FilesystemMigrationHelper) Migrate() {
+	if fmh.SourceURL == "" {
```
davcrypto, line 24: Move validation to `NewFilesystemMigrationHelper`. And could we change `SourceURL` and `DatabaseURL ` to private?

infrastructure/pg/migrationhelper/github/github.go
```diff
@@ -0,0 +1,86 @@
...
+) *GithubMigrationHelper {
+	return &GithubMigrationHelper{
+		SourceURL:   sourceURL,
+		DatabaseURL: databaeURL,
+	}
+}
+
+// Implement MigrationHelper interface
+func (gmh *GithubMigrationHelper) Migrate() {
+	if gmh.SourceURL == "" {
```
davcrypto, line 38: same for here

infrastructure/pg/migrationhelper/filesystem/filesystem.go
ysong42: sure, will change it!

infrastructure/pg/migrationhelper/github/github.go
ysong42: 👌

PR #640, opened 2021-12-08
Problem: Missing test cases in Bridge Activity projections

calvinaco:
Solution: Add test cases for Bridge Pending Activities projection

### DIFF
--- a/appinterface/projection/rdbprojectionbase/base.go
+++ b/appinterface/projection/rdbprojectionbase/base.go
@@ -29,8 +29,8 @@ func NewRDbBase(rdbHandle *rdb.Handle, projectionId string) *Base {
 		rdbHandle,
 		projectionId,
 		Options{
-			MaybeConfigPtr: nil,
-			MaybeTable:     primptr.String(DEFAULT_TABLE),
+			MaybeReadConfigPtr: nil,
+			MaybeTable:         primptr.String(DEFAULT_TABLE),
 		},
 	)
 }
@@ -45,7 +45,8 @@ func NewRDbBaseWithOptions(rdbHandle *rdb.Handle, projectionId string, options O
 		Base: projection_usecase.NewBaseWithOptions(
 			projectionId,
 			projection_usecase.Options{
-				MaybeConfigPtr: options.MaybeConfigPtr,
+				MaybeReadConfigPtr: options.MaybeReadConfigPtr,
+				MaybeConfigPtr:     options.MaybeConfigPtr,
 			},
 		),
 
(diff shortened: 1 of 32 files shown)


appinterface/projection/rdbprojectionbase/base.go
```diff
@@ -57,10 +58,14 @@ func NewRDbBaseWithOptions(rdbHandle *rdb.Handle, projectionId string, options O
 }
 
 type Options struct {
-	// Pointer to the configuration
+	// Pointer to the configuration to be read from Toml file
+	MaybeReadConfigPtr interface{}
```
davcrypto, line 62: our app should not specifically binding to Toml config file, may i know why we need `MaybeReadConfigPtr`?

ysong42: Same question here. After adding this, it seems `BridgeActivityMatcher.New()` will not read config from toml file.

calvinaco: The original behaviour of `MaybeConfigPtr` is to read from the global projection config which is reading from Toml file:
https://github.com/crypto-com/chain-indexing/pull/640/files#diff-2bf44a05f77f5d8ea19f90ed17e81e60731cff8b12fa4d096f98b61b3088bfe4R42

`GlobalConfig` is a singleton https://github.com/crypto-com/chain-indexing/blob/master/usecase/projection/config.go
read in app

calvinaco: For the test case I found it easier to inject the config directly is more readable, that's why I introduce a new way to provide the config.

In the process I found the original `MaybeConfigPtr` is also confusion, and hence I update the naming as well.

calvinaco: But I agree the assumption of it always being read from Toml may not hold in the future.

ysong42 (commented):
Looks good to me!

Just one small comment. In `BridgeActivityMatcher.New()`, `MaybeConfigPtr` should be updated to `MaybeReadConfigPtr`.

https://github.com/crypto-com/chain-indexing/blob/4acdfbd64dca45d9a431636eea67fa92d558c17f/projection/bridge_activity/bridge_activity_matcher/bridge_activity_matcher.go#L57

image

appinterface/projection/rdbprojectionbase/base.go
calvinaco: Opened a new issue to handle this issue. https://github.com/crypto-com/chain-indexing/issues/648

Will put this PR on hold and work on it first.

davcrypto:
LGTM!

davcrypto (approved):

PR #643, opened 2021-12-08
Problem: Bridge pending activity projection supports only one channel

### LINKED ISSUE #641: Problem: Bridge pending activity projection supports only one channel

calvinaco:
Solution: (Fix #641) Add multiple channel supports and test cases

### DIFF
--- a/projection/bridge_activity/bridge_pending_activity/bridge_pending_activity.go
+++ b/projection/bridge_activity/bridge_pending_activity/bridge_pending_activity.go
@@ -148,7 +148,7 @@ func (projection *BridgePendingActivity) HandleEvents(height int64, events []eve
 	for _, event := range events {
 		if msgIBCTransferTransfer, ok := event.(*event_usecase.MsgIBCTransferTransfer); ok {
 			channelId := msgIBCTransferTransfer.Params.SourceChannel
-			if !projection.isListenedChannel(channelId) {
+			if !projection.isListenedChannelAtBlockHeight(channelId, height) {
 				continue
 			}
 			counterpartyConfig := projection.mustGetCounterpartyChainConfigByListenedChannel(channelId)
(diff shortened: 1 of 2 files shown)


projection/bridge_activity/bridge_pending_activity/bridge_pending_activity.go
```diff
@@ -345,9 +361,11 @@ func (projection *BridgePendingActivity) HandleEvents(height int64, events []eve
 			}
 
 		} else if cronosSendToIBCCreatedEvent, ok := event.(*event_usecase.CronosSendToIBCCreated); ok {
```
davcrypto, line 363: Actaully are we able to remove this Cronos specific event because after this PR we will support multiple custom chain.

projection/bridge_activity/bridge_pending_activity/bridge_pending_activity.go
```diff
@@ -34,37 +36,44 @@ type BridgePendingActivity struct {
...
-	ChannelId             string `mapstructure:"channel_id"`
-	StartingHeight        int64  `mapstructure:"starting_height"`
+	ThisChainName      string                    `mapstructure:"this_chain_name"`
+	CounterPartyChains []CounterPartyChainConfig `mapstructure:"counteparty"`
+}
+
+type CounterPartyChainConfig struct {
+	ChainName      string `mapstructure:"chain_name"`
+	ChannelId      string `mapstructure:"channel_id"`
+	StartingHeight int64  `mapstructure:"starting_height"`
```
davcrypto, line 51: I think we missing the usage of this config, right?

ysong42 (approved):
LGTM!

projection/bridge_activity/bridge_pending_activity/bridge_pending_activity.go
calvinaco: Not exactly because this is a special type of bridge transaction on Cronos chain.

Some background story is that on Cronos you can think of it has the EVM layer and Cosmos layer.

Since most users are interacting with the EVM layer, so IBC transfers, which is Cosmos layer, is triggered by sending a ETH Tx. That's why we need this special event for detecting this kind of IBC transfers.

projection/bridge_activity/bridge_pending_activity/bridge_pending_activity.go
calvinaco: Good catch, updated the PR with the `StartingHeight` check

projection/bridge_activity/bridge_pending_activity/bridge_pending_activity.go
davcrypto: If we change to `ethereumSendToIBCCreatedEvent` or `evmSendToIBCCreatedEvent`, do you think it is a proper but generic naming?

PR #644, opened 2021-12-10
Problem: Lack of real time performance monitoring integration interface

### LINKED ISSUE #638: Problem: Lack of real time performance monitoring integration interface
Currently, indexing projection are lack of real time data/performance monitoring on application level. We could add some monitoring support to the project e.g. export Prometheus data

davcrypto:
Resolved #638 

- Added config for prometheus server
- Added prometheus data point for projection latest height and execution time

Example exported data:
Screenshot 2021-12-10 at 4 42 53 PM

### DIFF
--- a/bootstrap/app.go
+++ b/bootstrap/app.go
@@ -7,6 +7,7 @@ import (
 	"github.com/crypto-com/chain-indexing/appinterface/rdb"
 	projection_entity "github.com/crypto-com/chain-indexing/entity/projection"
 	applogger "github.com/crypto-com/chain-indexing/external/logger"
+	"github.com/crypto-com/chain-indexing/infrastructure/metric/prometheus"
 	"github.com/crypto-com/chain-indexing/infrastructure/pg"
 	"github.com/golang-migrate/migrate/v4"
 )
@@ -101,6 +102,9 @@ func (a *app) Run() {
 			if runErr := a.indexService.Run(); runErr != nil {
 				a.logger.Panicf("%v", runErr)
 			}
+		}()
+		go func() {
+			prometheus.Run(a.config.Prometheus.exportPath, a.config.Prometheus.port)
 		}()
 	}
 
--- a/bootstrap/app.go
+++ b/bootstrap/app.go
@@ -105,7 +105,9 @@ func (a *app) Run() {
 		}()
 		if a.config.Prometheus.Enable {
 			go func() {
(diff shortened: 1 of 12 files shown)


calvinaco (changes requested):
Awesome! Overall looks good to me. Just a tiny suggestion

bootstrap/app.go
```diff
@@ -102,6 +103,13 @@ func (a *app) Run() {
 				a.logger.Panicf("%v", runErr)
 			}
 		}()
+		if a.config.Prometheus.Enable {
+			go func() {
+				if runErr := prometheus.Run(a.config.Prometheus.ExportPath, a.config.Prometheus.Port); runErr != nil {
+					a.logger.Panicf("%v", runErr)
+				}
+			}()
+		}
```
ysong42, line 112: Shall we move the `prometheus` instance under `IndexService`?

Instead of directly importing `prometheus` to `bootstrap/syncmanager.go` and `entity/projection/manager.go`, I think we could do it through dependency injection.

ysong42 (commented):
Thanks! Overall looks good to me!

Just one question about shall we introduce `prometheus` through dependency injection.

bootstrap/app.go
davcrypto: I though about this as well, but that will mean anywhere we want to use `prometheus` will need many extra param passing. Let say in views, projections, parser or even outer package logic.

Or any example we will need DI for prometheus?

ysong42: There are a few reasons why I prefer to use dependency injection.

1. Readability. If not do it through dependency injection, it might confuse future developers. As they may spend more time to understand where does it initialize and how to use it.

2. Convention. I think `prometheus` is kind of like the existing logger. We already apply dependency injection for logger, maybe we should also do that for `prometheus`?

3. Unit Test. If doing through dependency injection, it will be easier for writing unit test. As we could make use of `interface`. If not do through dependency injection, it may be tricky to write unit test for it.

davcrypto: Understood, i will change to a package dependency injection.

davcrypto: Updated for a package dependency injection only. Since prometheus still better to keep it as an external package in case we can use in any parts of the app. Make it instance would result in unnecessary code complexity.  

For example, we would use it in the indexService, httpAPIServer, projections, views, parsers, or others external package.

ysong42 (approved):

calvinaco (approved):
LGTM

PR #649, opened 2021-12-18
Problem: Projection config is not flexible

### LINKED ISSUE #648: Problem: Projection config is not flexible
Right now projection config design is built-in to the projection base, but it also affect flexibility of offering configuration.

To align with the library design, I propose to have each projection accept a config and the reading should be done on the app side.

calvinaco:
Solution: (Fix #648) Refactor projection to accept config directly


ysong42 (approved):
LGTM!

davcrypto (approved):

davcrypto:
LGTM

PR #653, opened 2021-12-30
Problem: Missing clear setup documentation

### LINKED ISSUE #635: Problem: Missing clear setup documentation

davcrypto:
Resolved #635

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ func initRouteRegistry(
 ```
 
 ### Configuration
-```
+```go
 config := bootstrap.Config{
     FileConfig: bootstrap.FileConfig{
         Blockchain: bootstrap.BlockchainConfig{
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -62,7 +66,7 @@ func initRouteRegistry(
 ```
 
 ### Configuration
-```
+```go
```
ysong42, line 69: This  `Configuration` section is maybe a bit confusing to users. 
Maybe add more detail on its usage and which instance will consume it?

ysong42 (commented):
Thanks for the PR!

The code examples are super useful! But I think it is a bit difficult for users to follow.

I have a few suggestions:

1. Maybe provide a minimum directory structure in the beginning of the tutorial? So that users will know how to structure the project and the code examples easily.

1. Maybe we should add more description and instruction under each section? Like which path should the file be placed on.

1. Maybe create a separate file for the setup tutorial? Like `doc/initial-setup-tutorial.md`. And then place a link to that file in the root README.

README.md
davcrypto: I think the major purposes is to show what is the configuration about. Usually usage of configuration would not be provided since they are just 'configurations'

davcrypto:
Thanks for the review @ysong42 

For C1 and C3, these are more likely to show on the example implementation app and its readme doc. Generic tutorial would expected to have detail usage of each essential component. E.g. if custom parser are supported later, we should included some description as well.

For C2, we shouldn't restricted those things if other dev implement they own indexers base on this repo, thus there would be no suggested file structure /  for their own repo. If they really have no idea their file structure, they could refer to projections or other implementation under this repo.

ysong42 (approved):

PR #654, opened 2021-12-31
Problem: New validator projection. Port main logic. [Part 1]

ysong42:
The `view` and `migration` haven't been implemented yet. WIll implement in the following PR. This PR is mainly for giving a general view of the current design and progress.

Notes about `Validator.Status` (bonded/unbonding/unbonded). This field is useful when later calculating completion time for undelegation and redelegation.
```
// Notes on `Validator.Status`:
//
// - MsgCreateValidator, CreateGenesisValidator: Validator.Status = Unbonded
// - PowerChanged event:
//   - power > 0: Validator.Status = Bonded
//   - power = 0: Validator.Status = Unbonding, UnbondingValidator entry created
// - UnbondingValidator complete unbinding period: Validator.Status = Unbonded
```


projection/validator_delegation/utils/share.go
```diff
@@ -0,0 +1,29 @@
...
+
+	if validator.Shares.IsZero() {
+		issuedShares = amount.ToDec()
+	} else {
+
+		if validator.Tokens.IsZero() {
+			return coin.Dec{}, errors.New("validator tokens is zero, cannot issue new shares")
+		}
+
+		// newShares = existingShares * newTokens / existingTokens
```
davcrypto, line 24: Remove commented code

ysong42:
Hi @davcrypto @calvinaco, this PR is ready for review (finally)!

It mainly ports the data structures and logic related to delegations calculation. 

The implementation of migration scripts and views will be included in the next PR.

Thanks!

projection/validator_delegation/handle_msg.go
```diff
@@ -0,0 +1,329 @@
...
+	// Write the updated Validator to DB
+	if err := validatorsView.Update(validator); err != nil {
+		return fmt.Errorf("error validatorsView.Update(): %v", err)
+	}
+
+	return nil
+}
+
+// handleDelegate - Handle create/update delegation. Return newShares and error.
+func (projection *ValidatorDelegation) handleDelegate(
```
davcrypto, line 110: are this handler bond to `MsgDelegate` ? if yes, could we name it handleMsgDelegate?

projection/validator_delegation/handle_msg.go
```diff
@@ -0,0 +1,329 @@
...
+	validator.Tokens = validator.Tokens.Add(amount)
+	validator.Shares = validator.Shares.Add(newShares)
+	if err := validatorsView.Update(validator); err != nil {
+		return coin.ZeroDec(), fmt.Errorf("error validatorsView.Update(): %v", err)
+	}
+
+	return newShares, nil
+}
+
+func (projection *ValidatorDelegation) handleUndelegate(
```
davcrypto, line 175: same as above

projection/validator_delegation/handle_msg.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+
+	if err := projection.insertUBDQueue(rdbTxHandle, ubd, completionTime); err != nil {
+		return fmt.Errorf("error projection.insertUBDQueue(): %v", err)
+	}
+
+	return nil
+}
+
+func (projection *ValidatorDelegation) handleRedelegate(
```
davcrypto, line 225: same as above

projection/validator_delegation/handle_msg.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+
+	if err := projection.insertRedelegationQueue(rdbTxHandle, red, completionTime); err != nil {
+		return fmt.Errorf("error projection.insertRedelegationQueue(): %v", err)
+	}
+
+	return nil
+}
+
+func (projection *ValidatorDelegation) handleValidatorUnjailed(
```
davcrypto, line 303: same as above

projection/validator_delegation/handle_msg.go
ysong42: Not really, when handling `MsgCreateValidator`, it will also be invoked.

projection/validator_delegation/ubd_queue.go
```diff
@@ -0,0 +1,41 @@
+package validator_delegation
```
davcrypto, line 1: do ubd have a complete naming?

davcrypto:
Thanks @ysong42 for the PR! We will have complete abilities for the validator projection after this is done! Just a few comment:

- `handle_msg.go` expected to handle a message per function like `handle_xxx_events.go`  If this way doesn't make sense when doing implementation, could we use other way to group these function? e.g. handleDelegate to `delegation.go`
- `ubd_queue.go` change to `unbonding_delegation_queue.go`

projection/validator_delegation/handle_msg.go
ysong42: `VALIDATOR_JAILED` is an event introduced by ourselves. Not really an Msg.. Yea the file the is a bit misleading. I will revise that! Thanks!

ysong42:
@davcrypto Thanks for the comments. The PR is updated!

davcrypto (approved):

projection/validator_delegation/delegation.go
```diff
@@ -0,0 +1,269 @@
...
+	validatorsView := NewValidators(rdbTxHandle)
+	delegationsView := NewDelegations(rdbTxHandle)
+
+	// Get the validator
+	validator, found, err := validatorsView.FindByOperatorAddr(validatorAddress, height)
+	if err != nil {
+		return coin.ZeroDec(), fmt.Errorf("error validatorsView.FindByOperatorAddr(): %v", err)
+	}
+	if !found {
+		return coin.ZeroDec(), fmt.Errorf("error validator not found: %v, %v", validatorAddress, height)
```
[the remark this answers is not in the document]
ysong42: https://pkg.go.dev/fmt#hdr-Printing

Most of the time, I am just brainlessly using the `%v`, as it will print in the default format.

> %v	the value in a default format
	when printing structs, the plus flag (%+v) adds field names

projection/validator_delegation/delegation.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+
+	// Create/Get Delegation
+	delegation, found, err := delegationsView.FindBy(
+		validatorAddress,
+		delegatorAddress,
+		height,
+	)
+	if err != nil {
+		return coin.ZeroDec(), fmt.Errorf("error delegationsView.FindBy(): %v", err)
```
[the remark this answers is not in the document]
ysong42: I think if the method name changes in the future, developers should grep all the files and to see which parts need to be updated. In that case, maybe they should update the error message as well?

Even if they forget to update the error message, in the future, when there is an error return, users could still use the error messages to find out which call returns the error.

Another reason I did not introduce a more customized error here is, many of the errors here are just the wrapper of the original error. Providing a customized error seems a bit overkill...

projection/validator_delegation/delegation.go
```diff
@@ -0,0 +1,269 @@
...
+
+	// If the delegation belongs to operator of the validator and undelegating will decrease the validator's
+	// self-delegation below their minimum, we jail the validator.
+	if isOperatorDelegation && !validator.Jailed &&
+		delegationRemainingTokens.TruncateInt().LT(validator.MinSelfDelegation) {
+
+		if err := projection.jailValidator(rdbTxHandle, validator); err != nil {
+			return coin.ZeroInt(), fmt.Errorf("error in projection.jailValidator(): %v", err)
+		}
+		validator.Jailed = true
```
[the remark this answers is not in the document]
ysong42: So there are two separate fields store the `Status` (bonded/unbonding/unbonded) and `Jailed` (bool).

In this case, the `Validator.Jailed` will immediately be set to true. And there is a `PowerIndex` (in Cosmos), validator will be removed from the store immediately.
https://github.com/cosmos/cosmos-sdk/blob/9aef070625e9676d7c0acb212c17ae9dba3c32dc/x/staking/keeper/delegation.go#L675
https://github.com/cosmos/cosmos-sdk/blob/9aef070625e9676d7c0acb212c17ae9dba3c32dc/x/staking/keeper/val_state_change.go#L264

Then later, in the EndBlock section, the `Validator.Status` will be updated to `unbonding`.
 https://github.com/cosmos/cosmos-sdk/blob/9aef070625e9676d7c0acb212c17ae9dba3c32dc/x/staking/keeper/val_state_change.go#L112
 (check line 122, 128, 131, 183, 189, 194)
 
 To answer the question, after the block finishes everything, the `Validator.Status = unbonding` and `Validator.Jailed = true`.

projection/validator_delegation/delegation.go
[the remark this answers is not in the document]
ysong42: I re-think this and now I think you are right. Thanks for pointing this out. In other projects, seldom see the method name in the return errors. I will update it and use a more meaningful error message (will apply to other places as well)!

projection/validator_delegation/redelegation.go
```diff
@@ -0,0 +1,205 @@
...
+	// Get the validator
+	validator, found, err := validatorsView.FindByOperatorAddr(validatorSrcAddress, height)
+	if err != nil {
+		return completionTime, false, fmt.Errorf("error validatorsView.FindByOperatorAddr(): %v", err)
+	}
+
+	switch {
+	case !found || validator.IsBonded():
+		// the longest wait - just unbonding period from now
+		completionTime = blockTime.Add(projection.config.unbondingTime)
```
calvinaco, line 108: The unbonding time may changed on-chain. This is better to be retrieved from
- genesis
- monitor for governance proposal when it is updated

projection/validator_delegation/delegation.go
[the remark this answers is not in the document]
calvinaco: There's a subtle differences actually:
https://stackoverflow.com/questions/41159169/why-not-use-v-to-print-int-and-string

But in this case I think it should be fine as it's a string type.

usecase/event/evidence.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/crypto-com/chain-indexing/usecase/model"
+
+	jsoniter "github.com/json-iterator/go"
+
+	event_entity "github.com/crypto-com/chain-indexing/entity/event"
+	"github.com/luci/go-render/render"
+)
+
+const EVIDENCE = "Evidence"
```
calvinaco, line 14: I think we are lacking the very here. For example `EvidenceSubmitted`

usecase/command/evidence.go
```diff
@@ -0,0 +1,35 @@
...
+func NewEvidence(blockHeight int64, params model.EvidenceParams) *Evidence {
+	return &Evidence{
+		blockHeight,
+		params,
+	}
+}
+
+// Name returns name of command
+func (*Evidence) Name() string {
+	return "Evidence"
```
calvinaco, line 23: I think we are lacking the very here, for example `SubmitEvidence` or `CreateEvidence` etc

ysong42: Yeah, thanks! Good catch. WIll change it to `EvidenceSubmitted`.

projection/validator_delegation/handle_end_block.go
```diff
@@ -0,0 +1,180 @@
...
+	// Remove those mature UnbondingValidator entry
+	if err := unbondingValidatorsView.DeleteMatureEntries(blockTime); err != nil {
+		return fmt.Errorf("error unbondingValidatorsView.DeleteMatureEntries(): %v", err)
+	}
+
+	return nil
+}
+
+// Find mature UnbindingDelegation through UBDQueue and then handle them one by one
+func (projection *ValidatorDelegation) handleMatureUBDQueueEntries(
```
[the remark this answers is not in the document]
ysong42: Yes, I also find the UBD confusing when I first read this.

Maybe let's remove it completely and use the full name? In fact, I tried it before, and then I found those method names and variable names grow too long. So in the end I follow the Cosmos way.

Now David and you both raise this. I will change it to use the full name!

projection/validator_delegation/redelegation.go
ysong42: Good catch! Thanks for pointing this out!

projection/validator_delegation/delegation.go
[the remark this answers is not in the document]
calvinaco: Does this jail involve a jailing period? Or it's just a status update?

Because usually when we say a validator jailed it means the validator cannot perform any action during jailing period.

projection/validator_delegation/redelegation.go
calvinaco: We can do it on next PR if it is too large. But make sure we have a issue to track it. Up to you :)

ysong42: Yeah, I agree. I just checked it. Seems many changes need to be introduced. I prefer to do it in the next PR.

ysong42: https://github.com/crypto-com/chain-indexing/issues/663. Issue created.

projection/validator_delegation/delegation.go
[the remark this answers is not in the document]
ysong42: No, in this case, there is no jailing period. If the validator delegate enough tokens to itself, it can be immediately unjailed through a MsgUnjail.

Here is a good reference to the question: https://github.com/cosmos/cosmos-sdk/blob/2f335a2b8cc4fcfa337c104823fab9f6ecb387da/x/slashing/keeper/unjail.go#L40

I think the jailing period you mentioned relates to `ValidatorSigningInfo`, there is a `JailedUntil` field under it. But `JailedUntil`seems only in use when the jail is due to `missing_signature`.

`ValidatorSigningInfo` link: https://github.com/cosmos/cosmos-sdk/blob/2f335a2b8cc4fcfa337c104823fab9f6ecb387da/x/slashing/types/slashing.pb.go#L33

calvinaco (approved):
LGTM

PR #656, opened 2022-01-05
Problem: Missing example app

### LINKED ISSUE #634: Problem: Missing example app
One suggestion is to put both example app and the current codebase into this repository. 

\- repository root
|- lib, current codebase
|- app-example, an example application with basic API and leveraging the `lib`

davcrypto:
Resolved #634


example/app/example-app/projections.go
```diff
@@ -0,0 +1,327 @@
...
+	githubMigrationHelperConfig := github_migrationhelper.Config{
+		GithubAPIUser:    params.GithubAPIUser,
+		GithubAPIToken:   params.GithubAPIToken,
+		MigrationRepoRef: params.MigrationRepoRef,
+		ConnString:       connString,
+	}
+
+	switch name {
+	case "Example":
+		sourceURL := github_migrationhelper.GenerateDefaultSourceURL("Example", githubMigrationHelperConfig)
```
ysong42, line 119: The `sourceURL` is wrong here. Might need to handle custom projection in a different way.

Maybe refer to the code we used in `chain-indexing-server`: https://github.com/crypto-com/chain-indexing-server/blob/ee1e65dc5f0441eba6f07f71bf842e7de05e919f/app/chain-indexing-app/projections.go#L284

ysong42 (commented):
Thanks for the PR! I believe this will be super useful for our future users!

I spotted some small issues. And I cannot enable the `Example` projection, it is caused by the migration source URL generation logic. 

Thanks!

example/app/example-app/projections.go
davcrypto: Thanks and good catch! Updated custom GitHub path

example/Makefile
```diff
@@ -0,0 +1,5 @@
+build: chain-indexing-app
+chain-indexing-app:
+	go build ./app/example-app/
+clean:
+	rm example-app
```
calvinaco, line 5:
```suggestion
	rm -i example-app
```
how about add a prompt before removing?

calvinaco: Though not sure if it is violating the Makefile common practice

example/app/example-app/projections.go
```diff
@@ -0,0 +1,352 @@
...
+		MigrationRepoRef: config.GithubAPI.MigrationRepoRef,
+
+		ServerMigrationRepoRef: customConfig.ServerGithubAPI.MigrationRepoRef,
+	}
+	for _, projectionName := range config.Projection.Enables {
+		projection := InitProjection(
+			projectionName, initParams,
+		)
+		if projection == nil {
+			continue
```
calvinaco, line 77: See comment below on InitProjection behaviour comment

example/app/example-app/projections.go
```diff
@@ -0,0 +1,352 @@
...
+				)
+				continue
+			}
+		}
+		projections = append(projections, projection)
+	}
+
+	// Append additional projection
+	for _, projectionName := range config.Projection.Enables {
+		projection := InitAdditionalProjection(
```
calvinaco, line 92: similar as above

example/app/example-app/projections.go
```diff
@@ -0,0 +1,352 @@
...
+	switch name {
+	case "Example":
+		sourceURL := generateGithubMigrationSrouceURLForCustomProjection(name, githubMigrationHelperConfigForCustomProjection)
+		databaseURL := migrationhelper.GenerateDefaultDatabaseURL(name, connString)
+		migrationHelper := github_migrationhelper.NewGithubMigrationHelper(sourceURL, databaseURL)
+
+		return example.NewAdditionalProjection(params.Logger, params.RdbConn, migrationHelper)
+	}
+
+	return nil
```
calvinaco, line 133: I am thinking if we should panic here instead?

example/app/example-app/routes/routeregistry.go
```diff
@@ -0,0 +1,39 @@
+package routes
```
calvinaco, line 1: `routeregistry.go` instead of `routerregistry.go`?

example/Makefile
davcrypto: Shall be okay

calvinaco (changes requested):

example/app/example-app/run.go
```diff
@@ -0,0 +1,194 @@
...
+	applogger "github.com/crypto-com/chain-indexing/external/logger"
+	"github.com/crypto-com/chain-indexing/external/primptr"
+	"github.com/crypto-com/chain-indexing/infrastructure"
+	projection_usecase "github.com/crypto-com/chain-indexing/usecase/projection"
+)
+
+func run(args []string) error {
+	cliApp := &cli.App{
+		Name:                 filepath.Base(args[0]),
+		Usage:                "Crypto.com Chain Indexing Service",
```
calvinaco, line 22:
```suggestion
		Usage:                "Crypto.org Chain Indexing Service",
```

example/httpapi/handlers/example.go
```diff
@@ -0,0 +1,56 @@
...
+type Example struct {
+	logger applogger.Logger
+
+	exampleView example_view.Examples
+}
+
+func NewExample(logger applogger.Logger, rdbHandle *rdb.Handle) *Example {
+	return &Example{
+		logger.WithFields(applogger.LogFields{
+			"module": "DesktopWalletAnalytics",
```
calvinaco, line 22:
```suggestion
			"module": "ExampleModule",
```

example/app/example-app/routes/routeregistry.go
davcrypto: should be route registry

example/app/example-app/projections.go
davcrypto: Reason why not using panic here is that we separated projections initialisation function into two (provided one + additional one) and this way could make the customisation more clear for devs.

example/projection/example/example.go
```diff
@@ -0,0 +1,102 @@
...
+			_ = rdbTx.Rollback()
+		}
+	}()
+
+	rdbTxHandle := rdbTx.ToHandle()
+
+	examplesView := NewExamplesView(rdbTxHandle)
+
+	for _, event := range events {
+		log.Println(event)
```
calvinaco, line 76: Should use logger or remove if for debug purpose?

calvinaco (commented):
Overall looks good to me! Left some comment.

example/README.md
```diff
@@ -0,0 +1,95 @@
...
+        
+        // Containing functions to initialize projections
+        projections.go
+        
+        // CLI runner
+        run.go
+        
+        routes/
+            // Implementation for `Route` interface
+            routerregistry.go
```
calvinaco, line 39:
```suggestion
            routeregistry.go
```

To be updated here after updating file name.

example/projection/example/example.go
davcrypto: Good catch! Removed

example/app/example-app/projections.go
ysong42: I think panic is better here. As we can give a more explicit error message to the user, `Man, you input an undefined projection name` 😂  It could help the user save sometimes for debugging his own typo.

davcrypto: As devs using the config file provided in this example, they are not likely to have typo for projection names. If we use panic, it will mean we will combining the initialisations together. Not a very good way to show devs which part is customisation projection and which part is provided projection. Do you think it is still worth to do it?

ysong42: I see. Yea, that also make sense. I am ok with the current design.

davcrypto: @calvinaco do you agree?

ysong42 (approved):
LGTM!

example/app/example-app/projections.go
calvinaco: Yes I think as an example app this is fine.

In a more production setup I will prefer to use panic because typo will happens (as always haha :)) and I want to avoid that error being swallowed.

calvinaco:
LGTM!

PR #662, opened 2022-01-12
Problem: Missing generic configuration

### LINKED ISSUE #657: Problem: Missing generic configuration
Currently, configurations are highly specific e.g. toml tags and variable from env. We could make this to be more generic for variety of implementation ways.

davcrypto:
Resolved #657 , #623 and #648


example/app/example-app/projections.go
```diff
@@ -313,7 +315,13 @@ func InitProjection(name string, params InitProjectionParams) projection_entity.
...
 		migrationHelper := github_migrationhelper.NewGithubMigrationHelper(sourceURL, databaseURL)
 
-		return bridge_pending_activity.NewBridgePendingActivity(params.Logger, params.RdbConn, migrationHelper)
+		config := bridge_pending_activity.Config{}
+		err := config.Fill(params.ExtraConfigs[name])
+		if err != nil {
+			params.Logger.Panicf(err.Error())
+		}
+
+		return bridge_pending_activity.NewBridgePendingActivity(params.Logger, params.RdbConn, migrationHelper, bridge_pending_activity.Config{})
```
calvinaco, line 324: Are we supposed to use `config` here?

example/app/example-app/run.go
```diff
@@ -78,7 +78,7 @@ func run(args []string) error {
 
 			&cli.StringFlag{
 				Name:    "tendermintURL",
-				Usage:   "Tendermint HTTP RPC URL",
+				Usage:   "TendermintApp HTTP RPC URL",
```
calvinaco, line 81: Is this change on purpose?

example/app/example-app/cronjobs.go
```diff
@@ -58,7 +60,13 @@ func InitCronJob(name string, params InitCronJobParams) projection_entity.CronJo
 		databaseURL := migrationhelper.GenerateDefaultDatabaseURL(name, connString)
 		migrationHelper := github_migrationhelper.NewGithubMigrationHelper(sourceURL, databaseURL)
 
-		return bridge_activity_matcher.New(params.Logger, params.RdbConn, migrationHelper)
+		config := bridge_activity_matcher.Config{}
+		err := config.Fill(params.ExtraConfigs[name])
```
calvinaco, line 64: If the use of `Fill` is expected to be followed by first allocating an empt struct, how about make it a new method?

`bridge_activity_matcher.DecodeConfigFromInterface(params.ExtraConfigs[name])`

projection/bridge_activity/bridge_activity_matcher/bridge_activity_matcher.go
```diff
@@ -37,6 +38,27 @@ type Config struct {
...
+		return fmt.Errorf("error creating cronjob config decoder: %v", decoderErr)
+	}
+
+	if err := decoder.Decode(data); err != nil {
+		return fmt.Errorf("error decoding cronjob BridgeActivityMatcher config: %v", err)
+	}
+
+	return nil
+}
+
```
calvinaco, line 61: Love the way for each projection to be self-contained on config reading.

example/app/example-app/projections.go
davcrypto: Good Catch! Updated

example/app/example-app/run.go
davcrypto: Should be the IDE refactor effect. Changed back.

example/app/example-app/cronjobs.go
davcrypto: Thanks for the advice, Updated

ysong42 (approved):
This is so cool! LGTM!

PR #683, opened 2022-02-07
Problem: Bridge activity API is not dynamic based on supported chain

calvinaco:
Solution: Accept config for dynamic API query

### DIFF
--- a/infrastructure/httpapi/handlers/bridges.go
+++ b/infrastructure/httpapi/handlers/bridges.go
@@ -26,14 +26,14 @@ type Bridges struct {
 
 	bridgeActivitiesView bridge_activitiy_view.BridgeActivities
 	accountAddressPrefix string
-	config               BridgeConfig
+	config               BridgesConfig
 }
 
 func NewBridges(
 	logger applogger.Logger,
 	rdbHandle *rdb.Handle,
 	cosmosAddressPrefix string,
-	config BridgeConfig,
+	config BridgesConfig,
 ) *Bridges {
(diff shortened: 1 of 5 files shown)


infrastructure/httpapi/handlers/bridges.go
```diff
@@ -366,3 +396,24 @@ func parseStatus(value string) (*types.Status, error) {
 		return nil, errors.New("unrecognized status")
 	}
 }
+
+func DefaultCronosEVMAddressHookGenerator(accountAddressPrefix string) func(string) (string, error) {
```
davcrypto, line 400: Could we put this kind of address parser in a separated package? e.g. bridges/cronos, bridges/parsers/cronos

So that to start to make the bridge activity be more generic

davcrypto:
Thanks @calvinaco  and we will have general bridge activity projection after this update. Overall LGTM, just a comment on the package convention.

ysong42 (approved):
LGTM!

infrastructure/httpapi/handlers/bridges.go
calvinaco: good suggstion

calvinaco: Updated, I moved it under the `external` package, wdyt?

davcrypto: yup, lets do in `external` first

davcrypto (approved):

PR #690, opened 2022-02-16
Problem: URL values in routes are missing existence check

### LINKED ISSUE #689: Problem: URL values in routes are missing existence check
## Background

Some of the routes expect URL value (e.g. /api/v1/bridges/{network}/account/{account}/activities). However, the existence check for these values are not checked properly.

calvinaco:
Solution: (Fix #689) Add value existence check to all API handlers


ysong42 (approved):
LGTM!

PR #721, opened 2022-07-26
Add starting block height config

### LINKED ISSUE #671: Problem: Missing support for chain start height
## Background

Some blockchains may start the chain at arbitrary height (e.g. Cosmos Hub). Right now we only support start from height 1 (Genesis block then height 1)

## Task Details

Support starting sync at a custom height in configuration.

davcrypto:
Resolved #671

### DIFF
--- a/bootstrap/config/config.go
+++ b/bootstrap/config/config.go
@@ -17,6 +17,7 @@ type Config struct {
 
 type IndexService struct {
 	Enable                     bool                       `yaml:"enable" toml:"enable" xml:"enable" json:"enable,omitempty"`
+	StartingBlockHeight        int64                      `yaml:"starting_block_height" toml:"starting_block_height" xml:"starting_block_height" json:"starting_block_height,omitempty"`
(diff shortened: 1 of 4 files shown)


vincentysc (approved):

vincentysc:
Looks good for me. But Im getting a `makeslice: len out of range` from `window.go` line 85 when trying to skip some blocks.
https://github.com/crypto-com/chain-indexing/blob/master/usecase/syncstrategy/window.go

davcrypto:
Thanks @vincentysc, may i have more detail to reproduce it? e.g. config

vincentysc:
> Thanks @vincentysc, may i have more detail to reproduce it? e.g. config

No problem, Im using the local chain to test. So basically, the indexer `currentBlockHeight` is 100, the chain `currentBlockHeight` is 200 and I set the `starting_block_height` to 150 in the `chain-indexing-server` `config.yaml`.

PR #724, opened 2022-07-28
FIX: nullable reference

davcrypto:


usecase/parser/msg.go
```diff
@@ -1235,6 +1241,12 @@ func parseRawMsgGenericGrant(
 		MaybeGenericGrant: &rawMsg,
 	}
 
+	// Getting possible signer address from Msh
```
calvinaco, line 1250:
```suggestion
	// Getting possible signer address from Msg
```

usecase/parser/msg.go
```diff
@@ -152,6 +152,12 @@ func ParseBlockTxsMsgToCommands(
 func ParseMsgSend(
 	parserParams utils.CosmosParserParams,
 ) ([]command.Command, []string) {
+	// Getting possible signer address from Msh
```
calvinaco, line 155:
```suggestion
	// Getting possible signer address from Msg
```

usecase/parser/msg.go
```diff
@@ -152,6 +152,12 @@ func ParseBlockTxsMsgToCommands(
...
+	var possibleSignerAddresses []string
+	if parserParams.Msg["from_address"] != nil {
+		possibleSignerAddresses = append(possibleSignerAddresses, parserParams.Msg["from_address"].(string))
+	}
```
calvinaco, line 159: Do we need to cater invalid MsgSend here?

calvinaco (changes requested):

usecase/parser/msg.go
davcrypto: right, updated

usecase/parser/msg.go
```diff
@@ -152,6 +152,12 @@ func ParseBlockTxsMsgToCommands(
 func ParseMsgSend(
 	parserParams utils.CosmosParserParams,
 ) ([]command.Command, []string) {
+	// Getting possible signer address from Msg
+	var possibleSignerAddresses []string
+	if parserParams.Msg != nil && parserParams.Msg["from_address"] != nil {
```
calvinaco, line 157: I am actually wondering if `parserParams.Msg["from_address"]` is `nil`able

davcrypto: Thanks for remind

calvinaco (approved):

PR #725, opened 2022-07-29
FIX: nullable reference for all msgs

vincentysc:


usecase/parser/msg.go
```diff
@@ -176,14 +177,14 @@ func ParseMsgMultiSend(
 ) ([]command.Command, []string) {
 	rawInputs, _ := parserParams.Msg["inputs"].([]interface{})
```
davcrypto, line 178: no need for the invalid data check here?

usecase/parser/msg.go
```diff
@@ -176,14 +177,14 @@ func ParseMsgMultiSend(
...
 	inputs := make([]model.MsgMultiSendInput, 0, len(rawInputs))
-	addresses := make([]string, 0, len(rawInputs))
+	possibleSignerAddresses := make([]string, 0, len(rawInputs))
 	for _, rawInput := range rawInputs {
 		input, _ := rawInput.(map[string]interface{})
 		inputs = append(inputs, model.MsgMultiSendInput{
 			Address: input["address"].(string),
 			Amount:  tmcosmosutils.MustNewCoinsFromAmountInterface(input["coins"].([]interface{})),
 		})
-		addresses = append(addresses, input["address"].(string))
```
davcrypto, line 186: no need for the invalid data check here?

vincentysc: oh, I add it back now.

usecase/parser/msg.go
```diff
@@ -1289,22 +1494,30 @@ func ParseMsgRevoke(
 			RawMsgRevoke: rawMsg,
 		}
 
+		// Getting possible signer address from Msg
+		var possibleSignerAddresses []string
+		possibleSignerAddresses = append(possibleSignerAddresses, revokeParams.Granter)
```
davcrypto, line 1503: may i know where do we assign `Granter` to `revokeParams`?

vincentysc: The `RawMsgRevoke` type got `Granter`
https://github.com/crypto-com/chain-indexing/blob/e6d137d63ffe8b75710ea64b356798e1219b5a3b/usecase/parser/msg.go#L1497

davcrypto: looks like `revokeParams. Granter` must be `nil`?

vincentysc: so using `revokeParams.RawMsgRevoke.Granter` instead, right?

davcrypto: `revokeParams.RawMsgRevoke.Granter` make more sense. And i saw other parsers might have similar situation, can you help to double check?

vincentysc: yea, sure

davcrypto (approved):

PR #744, opened 2022-09-02
Problem: Parser is dependent on the chain codebase

### LINKED ISSUE #712: Problem: Parser is dependent on the chain codebase
Right now the Tx parser requires registration of coded, which means it depends on the chain codebase. 

However, if we want to support different Cosmos chains, it would require including multiple chain codebase into the repository and that would potentially break because each chain may use a different versions of Cosmos SDK and they have do some global registration.

This task aims to look for alternatives mean of Tx parser. One seems to be promising solutions is to use the Tendermint RESTful API `/cosmos/tx/v1beta1/txs`.

davcrypto:
Resolved #712


vincentysc (approved):

infrastructure/cosmosapp/httpclient.go
```diff
@@ -656,6 +657,86 @@ func (client *HTTPClient) ProposalTally(id string) (cosmosapp_interface.Tally, e
 	return tallyResp.Tally, nil
 }
 
+func (client *HTTPClient) Txs(hash string) (*model.Tx, error) {
```
calvinaco, line 660: I have a small question here. While the method is named by following the RESTful API path name, but the method of `Txs` seems to imply it can be used to query multiple transaction.

Do we want to keep follow the API path or we should give the method a more meaningful name.

calvinaco (approved):
I have a comment on the method naming but the PR LGTM.

infrastructure/cosmosapp/httpclient.go
davcrypto: I will suggest `Tx`

Issue #723, opened 2022-07-27
Problem: Missing raw transaction projection


davcrypto:
Currently, a raw transaction projection is missed for generic consumption propose

PR #746, opened 2022-09-08
Problem: Missing raw transaction projection

davcrypto:
Resolved: #723


vincentysc (approved):

projection/raw_transaction/raw_transaction.go
```diff
@@ -0,0 +1,208 @@
...
+						Type:           transactionCreatedEventSigner.MaybeKeyInfo.Type,
+						IsMultiSig:     transactionCreatedEventSigner.MaybeKeyInfo.IsMultiSig,
+						Pubkeys:        transactionCreatedEventSigner.MaybeKeyInfo.Pubkeys,
+						MaybeThreshold: transactionCreatedEventSigner.MaybeKeyInfo.MaybeThreshold,
+					}
+				}
+				signers = append(signers, transactionRowSigner)
+			}
+			tx.Signers = signers
+			txs = append(txs, tx)
```
calvinaco, line 128: I remember we have a conversation about signer info maybe missing in some cases. Can you shared again the conclusion? Is it that the signer always present?

projection/raw_transaction/raw_transaction_integration_test.go
```diff
@@ -0,0 +1,135 @@
+package raw_transaction_test
```
calvinaco, line 1: do we need this file? Seems it is copying from the block projection.

projection/raw_transaction/raw_transaction_integration_test.go
```diff
@@ -0,0 +1,135 @@
...
+	model_usecase "github.com/crypto-com/chain-indexing/usecase/model"
+)
+
+var _ = Describe("RawTransaction", func() {
+	It("should implement projection", func() {
+		fakeLogger := NewFakeLogger()
+		fakeRdbConn := rdb_test.NewFakeRDbConn()
+		var _ entity_projection.Projection = block.NewBlock(fakeLogger, fakeRdbConn, nil)
+	})
+
```
calvinaco, line 29: This is some old naive tests I have back in the days. A better way could be just have a line in the source code file:
```
var _ entity_projection.Projection = &RawTransaction{}
```

projection/raw_transaction/raw_transaction.go
davcrypto: empty signer would be exist here since we do not process any data inside the messages

Issue #735, opened 2022-08-17
Problem: Missing raw account transaction projection


davcrypto:
Currently, a raw account transaction projection is missed for generic consumption propose

PR #752, opened 2022-09-21
Problem: Missing raw account transaction projection

davcrypto:
Resolve: #735


vincentysc:
Maybe we need to add the `httpapi handler` for `/accounts/{address}/rawTransactions` ?

vincentysc (approved):

Issue #383, opened 2021-04-17
Problem: Account projection balance is relying on Cosmos real time data


calvinaco:
Should use `GenesisCreated` and `AccountTransfer` event to construct the balance

davcrypto:
Tired to get the balance via events in txs results but looks like some fee are not included for failed tx (but deducted from wallet)

ref: https://github.com/cosmos/cosmos-sdk/issues/10618

davcrypto:
# Account Usable Balance

Altering go deeper into cosmos-sdk and take some experimental implementation. Usable balance is proved can be only invoke inside the bank module. As a result, this balance can be successfully accumulated by combining the following means base on v0.44.5 [cosmos-sdk](https://github.com/cosmos/cosmos-sdk)

## Genesis block account balance
All initial balance stated inside genesis block
```
{
  "jsonrpc": "2.0",
  "id": -1,
  "result": {
    "genesis": {
      ...
      "app_state": {
        ...
        "bank": {
          "params": {
            "send_enabled": [],
            "default_send_enabled": true
          },
          "balances": [
            {
              "address": "tcro1s4ggq2zuzvwg5k8vnx2xfwtdm4cz6wtnfl4xx7",
              "coins": [
                {
                  "denom": "basetcro",
                  "amount": "840000000000000000"
                }
              ]
            },
            {
              "address": "tcro14fzksz5h72et4ssjtqpwsmhz6ysk6r4ngtg6zj",
              "coins": [
                {
                  "denom": "basetcro",
                  "amount": "840000000000000000"
                }
              ]
            },
            {
              "address": "tcro163tv59yzgeqcap8lrsa2r4zk580h8ddrpzvf4w",
              "coins": [
                {
                  "denom": "basetcro",
                  "amount": "840000000000000000"
                }
              ]
            }
          ],
          ...
        },
        ...
    }
  }
}
```

## Accumulate balance by events emitted

### Begin Block Events
Event emmited to here would not have any execution in term of the namk module. We can skip this part for usable balance.

### Block Events
Event emmited would include all actually actions taken for every tx inside the blocks. Events related to usable balance would be:
- `coin_spent`
	- Balance spent for some action
- `coin_received`
	- Balance received for some action
- `coinbase`
	- Balance minted
- `burn`
	- Balance burned

### Begin Block Events
[...]

PR #753, opened 2022-09-21
Problem: Missing raw account event projection

vincentysc:


projection/account_raw_event/account_raw_event.go
```diff
@@ -0,0 +1,530 @@
...
+
+	eventsView := NewAccountRawEvents(rdbTxHandle)
+	totalView := NewAccountRawEventsTotal(rdbTxHandle)
+
+	eventRows := make([]view.AccountRawEventRow, 0)
+
+	for _, event := range events {
+		if rawBlockCreatedEvent, ok := event.(*event_usecase.BlockRawEventCreated); ok {
+			accountRawEvent := utils.NewParsedTxsResultLogEvent(&rawBlockCreatedEvent.Data.Content)
+			if rawBlockCreatedEvent.Data.Type == "coin_spent" {
```
calvinaco, line 110: This list is so magnificent. I have some follow-up questions in mind:

1. Have you read @davcrypto findings on this before https://github.com/crypto-com/chain-indexing/issues/383 ? I think one caveat of relying on event is that some events are not emitted between versions.
2. I wonder if every message handling / chain logic happens emit some kinds of events. Maybe @vincentysc you can share you insight.

calvinaco (commented):
Thanks for the PR. I left some follow-up comments.

projection/account_raw_event/account_raw_event.go
davcrypto: i think there is no need to switch between `event type`, coz raw event should have no parsing on the event data (other than base64 decode)

vincentysc: oh, the switch case is mainly for extracting the related accounts.

vincentysc: or is there any better way to extract it?

davcrypto: can just assuming the event is related to a specific if there is any event value started with address prefix

vincentysc: cool, let me update it then~

projection/account_raw_event/account_raw_event.go
```diff
@@ -0,0 +1,155 @@
...
+	totalView := NewAccountRawEventsTotal(rdbTxHandle)
+
+	eventRows := make([]view.AccountRawEventRow, 0)
+
+	for _, event := range events {
+		if rawBlockCreatedEvent, ok := event.(*event_usecase.BlockRawEventCreated); ok {
+			accounts := []string{}
+			// append related account
+			for _, attribute := range rawBlockCreatedEvent.Data.Content.Attributes {
+				if strings.HasPrefix(attribute.Value, projection.accountAddressPrefix) {
```
calvinaco, line 111: I suggest to do a Cosmos address validness check too. After all having a attribute value prefixed with account address prefix, e.g. `cro` may not be that uncommon actually.

davcrypto (approved):

PR #763, opened 2022-10-07
Problem: Message addresses in upper case and lower case are distinguish

vincentysc:
Convert the address to lower case in msg parser and account related projections.


davcrypto (approved):
LGTM

PR #770, opened 2022-10-21
Add response log to Tendermint api request

vincentysc:


infrastructure/tendermint/httpclient.go
```diff
@@ -192,10 +196,12 @@ func (client *HTTPClient) request(method string, queryString ...string) (io.Read
 	}
 	rawResp, err := client.httpClient.Do(req)
 	if err != nil {
+		client.logger.Debugf("error requesting Tendermint %s endpoint: %v %v", url, err, rawResp.Body)
```
davcrypto, line 199: i think no need to log body here

vincentysc: removed

davcrypto (approved):

PR #773, opened 2022-10-25
Problem: Hard delete for burned crypto-org chain NFT data

vincentysc:
- add NFT `status` into table `view_nft_tokens` and `view_nft_messages` to indicate `Minted`/`Burned`
- soft delete the burned NFT related records in table `view_nft_tokens` and `view_nft_messages`
- insert burned NFT message when soft delete
- remove the record in `view_nft_token_total` and `view_nft_message_total`
- show non-burned NFT only for the NFT APIs


davcrypto (approved):
Thanks @vincentysc LGTM!

PR #778, opened 2022-10-31
Problem: Attention status conflict with Bonded status operation on validator projection

vincentysc:
Change the validator attention status to an attention flag without replacing the status `bonded`, `inactive` etc. 
validator API returns `attention` field with true/false


projection/validator/migrations/20221031152131_alter_view_validators_add_attention.up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE view_validators
+    ADD attention BOOL NOT NULL DEFAULT FALSE;
+
+CREATE INDEX view_validators_attention_btree_index ON view_validators USING btree (attention);
```
davcrypto, line 4: May i know why this flag needs to be indexed?

vincentysc: Iriginally, I think we will filter by this flag, but seems like it is not the case. I will remove it~

davcrypto (approved):

PR #781, opened 2022-11-01
Fix: resolve duplicate NFT tokens with upsert

vincentysc:
The same `denom_id` and `token_id`on burned NFT will be used for minting a new NFT. To avoid the database unique key conflict, we use `upsert` NFT token ( update with condition `burned = true`) instead of `insert` on `MsgNFTMintNFT` event.


projection/nft/nft.go
```diff
@@ -445,7 +445,7 @@ func (projection *NFT) updateToken(
 func (projection *NFT) insertToken(
```
davcrypto, line 445: Please help to change this name as well

davcrypto (approved):

PR #838, opened 2023-05-23
Show burned nft messages

vincentysc:


infrastructure/httpapi/handlers/nfts.go
```diff
@@ -486,7 +492,11 @@ func (handler *NFTs) ListMessagesByToken(ctx *fasthttp.RequestCtx) {
 		return
 	}
 
-	httpapi.SuccessWithPagination(ctx, blocks, paginationResult)
```
davcrypto, line 489: Can we do the correction on the SQL rather hijack the total record number here?

davcrypto: Re-write the function inside `WithCustomTotalQueryFn`

davcrypto (approved):

